PR #37, opened 2024-12-17
feat: updated redirection for incorporation applications via magic link

JazzarKarim:
*Issue #:* https://github.com/bcgov/entity/issues/24226

This is a continuation of: https://github.com/bcgov/sbc-auth/pull/3179

Here's how it works:

https://github.com/user-attachments/assets/89887d26-76b0-4181-a6f2-3c75afbe9966

If the above video doesn't work, maybe try this (.webm extension):

[screen-capture (18).webm](https://github.com/user-attachments/assets/6769308c-1dc9-49e6-ab7e-dabe870ef043)

### DIFF
--- a/business-registry-dashboard/app/locales/en-CA.ts
+++ b/business-registry-dashboard/app/locales/en-CA.ts
@@ -175,21 +175,9 @@ export default {
       title: 'Business Already Added',
       description: 'The business {identifier} is already in your Business Registry List.'
     },
-    magicLinkMissingContactInfo: {
-      title: 'Invalid Link',
-      description: 'The link is missing contact information. Please provide an email or phone number.'
-    },
-    magicLinkCreateBusinessFailed: {
-      title: 'Error Creating Business',
-      description: 'An error occurred while creating the business.'
-    },
-    magicLinkGenericError: {
+    magicLinkIncorporateNowError: {
       title: 'Error Processing Link',
-      description: 'An error occurred while processing the link.'
-    },
-    magicLinkInvalidNameRequest: {
-      title: 'Invalid Name Request',
-      description: 'An error occurred while processing the name request. Please make sure the name request is affiliated with your account and try again.'
+      description: 'The link appears to be entered incorrectly. Please go back and click the full link.'
     }
   },
   entityAlertTypes: {
--- a/business-registry-dashboard/app/locales/en-CA.ts
+++ b/business-registry-dashboard/app/locales/en-CA.ts
@@ -186,6 +186,10 @@ export default {
     magicLinkGenericError: {
       title: 'Error Processing Link',
       description: 'An error occurred while processing the link.'
+    },
+    magicLinkInvalidNameRequest: {
+      title: 'Invalid Name Request',
+      description: 'An error occurred while processing the name request. Please make sure the name request is affiliated with your account and try again.'
     }
   },
   entityAlertTypes: {
--- a/business-registry-dashboard/app/locales/en-CA.ts
+++ b/business-registry-dashboard/app/locales/en-CA.ts
@@ -174,6 +174,18 @@ export default {
     magicLinkAlreadyAdded: {
       title: 'Business Already Added',
       description: 'The business {identifier} is already in your Business Registry List.'
+    },
+    magicLinkMissingContactInfo: {
+      title: 'Invalid Link',
+      description: 'The link is missing contact information. Please provide an email or phone number.'
+    },
+    magicLinkCreateBusinessFailed: {
+      title: 'Error Creating Business',
+      description: 'An error occurred while creating the business.'
+    },
+    magicLinkGenericError: {
+      title: 'Error Processing Link',
+      description: 'An error occurred while processing the link.'
     }
   },
   entityAlertTypes: {
--- a/business-registry-dashboard/app/middleware/00.magic-link.global.ts
+++ b/business-registry-dashboard/app/middleware/00.magic-link.global.ts
@@ -1,6 +1,6 @@
 export default defineNuxtRouteMiddleware((to) => {
   // Check if the route is an affiliation invitation link without locale prefix
-  if (to.path.startsWith('/affiliationInvitation/acceptToken') && !to.path.startsWith('/en-CA/')) {
+  if ((to.path.startsWith('/affiliationInvitation/acceptToken') || to.path.startsWith('/magicLink')) && !to.path.startsWith('/en-CA/')) {
     // Convert query params to URLSearchParams to maintain all existing query parameters
     const searchParams = new URLSearchParams(to.query as Record<string, string>)
     const queryString = searchParams.toString()
--- a/business-registry-dashboard/app/middleware/00.magic-link.global.ts
+++ b/business-registry-dashboard/app/middleware/00.magic-link.global.ts
@@ -1,6 +1,6 @@
 export default defineNuxtRouteMiddleware((to) => {
   // Check if the route is an affiliation invitation link without locale prefix
-  if ((to.path.startsWith('/affiliationInvitation/acceptToken') || to.path.startsWith('/incorporateNowFromNr')) && !to.path.startsWith('/en-CA/')) {
+  if ((to.path.startsWith('/affiliationInvitation/acceptToken') || to.path.startsWith('/incorporateNow')) && !to.path.startsWith('/en-CA/')) {
     // Convert query params to URLSearchParams to maintain all existing query parameters
     const searchParams = new URLSearchParams(to.query as Record<string, string>)
     const queryString = searchParams.toString()
--- a/business-registry-dashboard/app/middleware/00.magic-link.global.ts
+++ b/business-registry-dashboard/app/middleware/00.magic-link.global.ts
@@ -1,6 +1,6 @@
 export default defineNuxtRouteMiddleware((to) => {
   // Check if the route is an affiliation invitation link without locale prefix
-  if ((to.path.startsWith('/affiliationInvitation/acceptToken') || to.path.startsWith('/magicLink')) && !to.path.startsWith('/en-CA/')) {
+  if ((to.path.startsWith('/affiliationInvitation/acceptToken') || to.path.startsWith('/incorporateNowFromNr')) && !to.path.startsWith('/en-CA/')) {
     // Convert query params to URLSearchParams to maintain all existing query parameters
     const searchParams = new URLSearchParams(to.query as Record<string, string>)
     const queryString = searchParams.toString()
--- a/business-registry-dashboard/app/pages/incorporateNow.vue
+++ b/business-registry-dashboard/app/pages/incorporateNow.vue
@@ -5,11 +5,11 @@ import {
 } from '@bcrs-shared-components/enums'
 
 const { t } = useI18n()
-const brdModal = useBrdModals()
-const route = useRoute()
 const affNav = useAffiliationNavigation()
-const isLoading = ref(true)
 const affStore = useAffiliationsStore()
+const brdModal = useBrdModals()
+const isLoading = ref(true)
+const route = useRoute()
 
 // Define eligible corporation types for incorporation
 const ELIGIBLE_CORP_TYPES = [
@@ -40,29 +40,27 @@ const createBusinessFromNR = async (business: any) => {
 /**
  * Magic Link Processing Flow:
  * 1. Verify user authentication (done in dashboard layout)
- * 2. Extract and validate URL parameters (nrId, email, phone)
+ * 2. Extract and validate URL parameters (nr, email, phone)
  * 3. Fetch name request details
- * 4. Verify business eligibility
- * 5. Create business and redirect to dashboard
+ * 4. Create affiliation if not already affiliated
+ * 5. Verify business eligibility
+ * 6. Create business and redirect to dashboard if needed
  */
 onMounted(async () => {
   try {
     isLoading.value = true
     // Extract query parameters from the URL
-    const { nrId, email, phone } = route.query as { nrId: string, email: string, phone: string }
+    const { nr, email, phone } = route.query as { nr: string, email: string, phone: string }
 
-    // Ensure either email or phone is provided for contact
+    // Ensure either email or phone is provided as contact information
     if (!email && !phone) {
-      brdModal.openMagicLinkModal(
-        t('error.magicLinkMissingContactInfo.title'),
-        t('error.magicLinkMissingContactInfo.description')
-      )
-      return
+      throw new Error('Missing contact information')
     }
 
-    // Fetch name request details using provided credentials
-    const nameRequest = await fetchNameRequest(nrId, phone, email)
+    // Fetch and validate name request details
+    const nameRequest = await fetchNameRequest(nr, phone, email)
 
+    // Transform name request data to match expected business structure
     const business = {
       corpType: { code: nameRequest.legalType },
       nameRequest: {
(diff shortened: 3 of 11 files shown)


bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-37-8gkpatut.web.app

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-37-8gkpatut.web.app

severinbeauvais:
> You can try (while logged out or logged in), this link: https://business-registry-dev--pr-37-8gkpatut.web.app/magicLink?nrId=NR+0521856&phone=1112223333

Very cool! A couple of observations:
- There are several redirects and none of the BRD pages have a spinner or other indication that they're not the final destination. Can anything be done about this?
- I can login but I don't have a chance to set which account I'm using, so if I wasn't logged in then the draft IA will be created in the default (first) account under my login. Can you escalate this and get approval that this is OK? Maybe ask Olga and Jacqueline.

business-registry-dashboard/app/pages/magicLink.vue
```diff
@@ -0,0 +1,112 @@
...
+    const business = {
+      corpType: { code: nameRequest.legalType },
+      nameRequest: {
+        ...nameRequest,
+        nrNumber: nameRequest.nrNum, // Transform nrNum to nrNumber
+        requestActionCd: nameRequest.request_action_cd // Transform request_action_cd to requestActionCd
+      }
+    }
+
+    // Verify business eligibility and proceed with incorporation
```
severinbeauvais, line 80: Do you think the NR should be affiliated to the current account first?

business-registry-dashboard/app/pages/magicLink.vue
```diff
@@ -0,0 +1,112 @@
...
+const { isAuthenticated } = useKeycloak()
+
+// Define eligible corporation types for incorporation
+const ELIGIBLE_CORP_TYPES = [
+  CorpTypes.BENEFIT_COMPANY,
+  CorpTypes.COOP,
+  CorpTypes.BC_CCC,
+  CorpTypes.BC_COMPANY,
+  CorpTypes.BC_ULC_COMPANY
+]
```
severinbeauvais, line 20: Hmm, about Coops...

Coops are LEAR-only, but I wonder if users will get an email for them with a magic link. I don't see any reason why not. Check with Kevin. For this reason, I think they can be left here.

PS In the future, there are 4 continued-in entity types that will be added. But not for Directed Launch.

business-registry-dashboard/app/middleware/00.magic-link.global.ts
```diff
@@ -1,6 +1,6 @@
 export default defineNuxtRouteMiddleware((to) => {
   // Check if the route is an affiliation invitation link without locale prefix
-  if (to.path.startsWith('/affiliationInvitation/acceptToken') && !to.path.startsWith('/en-CA/')) {
+  if ((to.path.startsWith('/affiliationInvitation/acceptToken') || to.path.startsWith('/magicLink')) && !to.path.startsWith('/en-CA/')) {
```
severinbeauvais, line 3: I recommend naming this path according to its specific use. It is not a generic "magic link". It is specific to creating an Incorporation Application from a specified NR.

business-registry-dashboard/app/pages/magicLink.vue
```diff
@@ -0,0 +1,112 @@
...
+    if (!email && !phone) {
+      brdModal.openMagicLinkModal(
+        t('error.magicLinkMissingContactInfo.title'),
+        t('error.magicLinkMissingContactInfo.description')
+      )
+      return
+    }
+
+    // Fetch name request details using provided credentials
+    const nameRequest = await fetchNameRequest(nrId, phone, email)
```
severinbeauvais, line 69: FYI, another way to handle exceptions in an async call is like this:
```
const nameRequest = await fetchNameRequest(nrId, phone, email).catch(error => /* handle error here */)
```

Sometimes instead of handling the error I just return null (`catch(() => null)`, and the next line of code could verify that nameRequest isn't falsy.

business-registry-dashboard/app/pages/magicLink.vue
```diff
@@ -47,9 +47,9 @@ const createBusinessFromNR = async (business: any) => {
 onMounted(async () => {
   // Redirect unauthenticated users to login page with current URL as redirect target
   if (!isAuthenticated.value) {
-    const authWebUrl = useRuntimeConfig().public.authWebUrl
+    const registryHomeURL = useRuntimeConfig().public.registryHomeURL
     const redirectUrl = encodeURIComponent(window.location.href)
-    window.location.href = `${authWebUrl}/login?redirect=${redirectUrl}`
+    window.location.href = `${registryHomeURL}/login/?return=${redirectUrl}`
```
JazzarKarim, line 52: Special thanks to @seeker25 for spending some time with me to figure this out.

FYI, the code I had in auth web has been reverted. We didn't need any code changes in the end.

JazzarKarim:
> > You can try (while logged out or logged in), this link: https://business-registry-dev--pr-37-8gkpatut.web.app/magicLink?nrId=NR+0521856&phone=1112223333
> 
> Very cool! A couple of observations:
> 
> * There are several redirects and none of the BRD pages have a spinner or other indication that they're not the final destination. Can anything be done about this?
> * I can login but I don't have a chance to set which account I'm using, so if I wasn't logged in then the draft IA will be created in the default (first) account under my login. Can you escalate this and get approval that this is OK? Maybe ask Olga and Jacqueline.

Number 1 should be fixed now. You are absolutely right. Now it should be much much cleaner as I added spinners appropriately (before redirection and after redirection).

I'll following up in regards to number 2.

JazzarKarim:
> > Temporary Url for review: https://business-registry-dev--pr-37-8gkpatut.web.app
> 
> I thought your code would have redirected me for login if I wasn't already logged in, but it still shows an empty table. Easy to fix?

This should be fixed now. I moved my authentication code to the common layout. So, it will be applied now this the main route.

business-registry-dashboard/app/locales/en-CA.ts
```diff
@@ -186,6 +186,10 @@ export default {
     magicLinkGenericError: {
       title: 'Error Processing Link',
       description: 'An error occurred while processing the link.'
+    },
+    magicLinkInvalidNameRequest: {
+      title: 'Invalid Name Request',
+      description: 'An error occurred while processing the name request. Please make sure the name request is affiliated with your account and try again.'
```
severinbeauvais, line 192: We want the magic link to affiliate the NR automatically.

business-registry-dashboard/app/pages/magicLink.vue
JazzarKarim: I think it's a super good idea to add that. So, I added a check to make sure the NR is affiliated to the account. This prevents now from let's say, BCREG0013 clicking on the magic link and it actually being processed when the NR is actually in BCREG0020...

business-registry-dashboard/app/pages/incorporateNowFromNr.vue
```diff
@@ -77,9 +72,14 @@ onMounted(async () => {
       }
     }
 
+    // Load user's affiliated businesses and check if this name request is already affiliated
+    // This prevents unauthorized access to name requests that belong to other users
```
severinbeauvais, line 76: I think NRs can be affiliated to multiple accounts, so this doesn't prevent that.

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-37-8gkpatut.web.app

business-registry-dashboard/app/middleware/00.magic-link.global.ts
JazzarKarim: I changed the route's name to incorporateNowFromNr. So, the link would be:
https://business-registry-dev--pr-37-8gkpatut.web.app/magicLink?incorporateNowFromNr=NR+0521856&phone=1112223333

What do you think? Let me know so that I can follow up with Kevin so that he can change this in his PR as well.

business-registry-dashboard/app/pages/incorporateNowFromNr.vue
```diff
@@ -77,9 +72,14 @@ onMounted(async () => {
...
+    await affStore.loadAffiliations()
+    const isAffiliated = isNameRequestAffiliated(affStore.affiliations.results, business.nameRequest?.nrNumber)
```
severinbeauvais, line 78: I think we want the NR to be affiliated as part of the magic link flow.

This flow is analogous to the "incorporate now" flow from an approved NR in Namerequest. Re: https://github.com/bcgov/namerequest/blob/main/app/src/mixins/nr-affiliation-mixin.ts

severinbeauvais: By the way, when I tested the Namerequest flow, I deleted the draft IA and was left with the NR in my MBR. I think we want this behaviour here, too.

JazzarKarim: That'll be fixed here: https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/24885

business-registry-dashboard/app/pages/magicLink.vue
severinbeauvais: Oops, I meant that your code should affiliate the NR before creating the draft IA.

business-registry-dashboard/app/middleware/00.magic-link.global.ts
severinbeauvais: How about just "incorporateNow" since this is functionally the same as clicking that button in Namerequest UI 😀

severinbeauvais: Your sample link isn't what I had in mind and it's not working (404).

I thought the magic link would be:https://business-registry-dev--pr-37-8gkpatut.web.app/incorporateNowFrom?nr=NR+0521856&phone=1112223333

business-registry-dashboard/app/pages/incorporateNowFromNr.vue
severinbeauvais: Sorry, I'm saying something else.

I'm saying that Namerequest UI affilaites the NR first, and then creates a draft IA, so that if the user deletes the draft IA, they are left with the NR in their MBR. This is good. Ideally the magic incorporation link process would work the same way.

business-registry-dashboard/app/middleware/00.magic-link.global.ts
JazzarKarim: I edited it, sorry I had it in the wrong format. It should work now. Either way, I'm changing it to incorporateNow.

severinbeauvais: I just tested it. Looks great!

severinbeauvais: Now I'm testing error scenarios.

Did you know this error appears if I leave out the NR number? (https://business-registry-dev--pr-37-8gkpatut.web.app/en-CA/incorporateNowFromNr)

image

JazzarKarim: Yep, this is because the first check the code makes is:
image

I can probably handle this in a more elegant way/dialog though. I'll have a look.

severinbeauvais: Or handle all 3 cases with the same dialog that simply tells them the link is invalid.

I hesitated whether to mention this, but the error isn't going to help any user if they see it (*). It's also a bit technical and we might not want to publicize exactly how this works.

(*) How about "The link is invalid. Please click the link from its original page." or something like that? I'm trying to make this useful for the user while imagining how they might get an invalid link and giving them some useful advice.

Or ask Jacqueline 😄

JazzarKarim: Had a call with Jacqueline about this Sev, we're gonna handle all 3 cases via the following:
"The link appears to be entered incorrectly. Please go back and click the full link."

business-registry-dashboard/app/locales/en-CA.ts
JazzarKarim: Should be fixed now.

business-registry-dashboard/app/pages/incorporateNow.vue
```diff
@@ -72,34 +70,41 @@ onMounted(async () => {
...
-      } catch (error) {
-        console.error('Error creating named business:', error)
-        brdModal.openMagicLinkModal(
-          t('error.magicLinkCreateBusinessFailed.title'),
-          t('error.magicLinkCreateBusinessFailed.description')
-        )
-      }
+    // Create affiliation if it doesn't exist
+    if (!isAffiliated) {
+      await affStore.createNRAffiliation({
```
JazzarKarim, line 79: Creating the affiliation here (if not already affiliated).

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-37-8gkpatut.web.app

business-registry-dashboard/app/pages/incorporateNow.vue
```diff
@@ -72,34 +70,41 @@ onMounted(async () => {
...
+        businessIdentifier: business.nameRequest?.nrNumber,
+        ...(phone && { phone }),
+        ...(email && { email })
+      })
+    }
+
+    // Get the temporary business identifier if it exists
+    const tempBusinessIdentifier = getTempBusinessIdentifierOfNameRequest(affStore.affiliations.results, business.nameRequest?.nrNumber)
```
JazzarKarim, line 87: I noticed a bug. If the user clicks on the link again and again, a new draft is created every time. This now, prevents that. IF a draft IA has been created before, we will direct the user to the previously created draft temp business identifier.

business-registry-dashboard/app/utils/affiliations.ts
```diff
@@ -284,3 +284,29 @@ export function isBusinessAffiliated (affiliations: Business[], businessIdentifi
...
+ * Gets the temporary business identifier associated with a Name Request
+ * @param affiliations - Array of affiliated businesses
+ * @param nrNumber - Name Request number to look up
+ * @returns The business identifier if found, undefined otherwise
+ */
+export function getTempBusinessIdentifierOfNameRequest (affiliations: Business[], nrNumber: string): string | undefined {
+  if (isNameRequestAffiliated(affiliations, nrNumber)) {
+    return affiliations.find(business => business.nameRequest?.nrNumber === nrNumber)?.businessIdentifier
+  }
+  return undefined
```
severinbeauvais, line 311: Would this work just as well if you returned null? I'm wary of using undefined because an variable with a value of undefined is not reactive.

business-registry-dashboard/app/pages/incorporateNow.vue
severinbeauvais: Great idea.

Future: check the Auth db whether any business exists for this IA (across the board). This would prevent another account from affiliating the NR and creating a second IA from it.

business-registry-dashboard/app/pages/incorporateNow.vue
```diff
@@ -72,34 +70,41 @@ onMounted(async () => {
...
+
+    // Check basic eligibility conditions
+    const isEligibleRequest = business.nameRequest.requestActionCd === NrRequestActionCodes.NEW_BUSINESS &&
+      ELIGIBLE_CORP_TYPES.includes(business.nameRequest.legalType)
+
+    if (!isEligibleRequest) {
+      throw new Error('Invalid name request')
+    }
```
severinbeauvais, line 95: I recommend moving these checks before the affiliation above.

business-registry-dashboard/app/pages/incorporateNow.vue
severinbeauvais, line 1: So we now have 3 places where a NR can be turned into an IA:
- Namerequest UI
- BRD under the Incorporate Now button
- BRD under this magic link

I'm thinking, in a future ticket, we could combine the last 2 and redirect the first one to here instead. Thoughts? Please create a new ticket if you agree.

business-registry-dashboard/app/pages/magicLink.vue
severinbeauvais: Looks good.

business-registry-dashboard/app/middleware/00.magic-link.global.ts
severinbeauvais: Fantastic! Love it.

business-registry-dashboard/app/pages/incorporateNowFromNr.vue
severinbeauvais: New flow looks great. And thanks for the super-useful comment in the code.

business-registry-dashboard/app/pages/incorporateNow.vue
severinbeauvais: In general, do the checks in a order that will fail fastest for the user.

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-37-8gkpatut.web.app

severinbeauvais (approved):
This works great! Looks good, too. Approved (even you still change a few small things).

business-registry-dashboard/app/pages/incorporateNow.vue
JazzarKarim: Agreed. New ticket: https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/25005

business-registry-dashboard/app/pages/incorporateNow.vue
```diff
@@ -70,29 +70,37 @@ onMounted(async () => {
       }
     }
 
+    // Before the affiliation checks, add a guard clause
+    if (!business.nameRequest?.nrNumber) {
```
JazzarKarim, line 74: Added a guard clause here just to fix an issue that typescript was complaining about.

business-registry-dashboard/app/pages/incorporateNow.vue
JazzarKarim: So you're thinking that on button press of "Register Now", we would take the user to this new route and let the route do it's thing? This way both cases would work in the same way instead of having different code for both cases. Hmmm, that sounds about right to me. I'll create a ticket for that.

JazzarKarim: New ticket: https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/25008

meawong (approved):
Tested and LGTM - awesome work!

meawong (approved):

ArwenQin (approved):
Tested and Looks all good!

PR #40, opened 2024-12-19
fix: fixed no affiliation records showing before load

JazzarKarim:
Ticket #: https://github.com/bcgov/entity/issues/24979

Fixed no affiliation records showing on initial page load which is wrong since it's showing before the affiliations are actually loaded.

### DIFF
--- a/business-registry-dashboard/app/stores/affiliations.ts
+++ b/business-registry-dashboard/app/stores/affiliations.ts
@@ -17,7 +17,7 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () => 
       type: '',
       status: ''
     },
-    loading: false,
+    loading: true,
     results: [] as Business[],
     count: 0
   })
--- a/business-registry-dashboard/package.json
+++ b/business-registry-dashboard/package.json
@@ -2,7 +2,7 @@
   "name": "business-registry-dashboard",
   "private": true,
   "type": "module",
-  "version": "0.0.8",
+  "version": "0.0.9",
   "scripts": {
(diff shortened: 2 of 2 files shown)


business-registry-dashboard/app/stores/affiliations.ts
```diff
@@ -17,7 +17,7 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () =>
       type: '',
       status: ''
     },
-    loading: false,
+    loading: true,
```
JazzarKarim, line 20: I figured this would be the smartest and cleanest way to do it with minimal code change. Now, the initial value for loading is true so it doesn't go from: Not Loading --> Loading --> Not Loading (finished).

We now go: Loading --> Not Loading (finished)

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-40-0vx1zsek.web.app

meawong (approved):
LGTM

business-registry-dashboard/app/stores/affiliations.ts
severinbeauvais: Awesome!

Does "no affiliation records" still display after loading (if needed)?

severinbeauvais (approved):

ArwenQin (approved):
LGTM!

business-registry-dashboard/app/stores/affiliations.ts
JazzarKarim: Yup, it does! This is an account with 0 items in the table:
image

PR #43, opened 2025-01-02
fix: existing businesses route to business dashboard (removed FF check)

JazzarKarim:
https://github.com/bcgov/entity/issues/25076

NRs for entity types not in the FF redirect to COLIN; left as is.
Existing businesses route now to Business Dashboard -- don't check FF.

Basically:
If type is NR, we are checking the `ia-supported-entities-brd` FF. If the FF doesn't support the type, redirect to COLIN.

For existing businesses though, we no more check the FF. We will ALWAYS go to the business dashboard with that business.


bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-43-avxs2fcd.web.app

severinbeauvais (approved):
LGTM

ketaki-deodhar (approved):

meawong (approved):

PR #44, opened 2025-01-03
fix: fixed the registries search endpoint

JazzarKarim:
Ticket: https://github.com/bcgov/entity/issues/24879

Moved to use the v1 endpoint (just like the old MBR).


business-registry-dashboard/app/utils/reg-search.ts
```diff
@@ -4,23 +4,16 @@ export async function regSearch (queryStr: string): Promise<RegSearchResult[]> {
...
   const token = await keycloak.getToken()
 
-  const response = await $fetch<RegSearchResponse>(`${config.regSearchApiUrl}/search/businesses`, {
-    method: 'POST',
+  const legalType = 'A,BC,BEN,C,CBEN,CC,CCC,CP,CUL,FI,GP,LL,LLC,LP,PA,S,SP,ULC,XCP,XL,XP,XS'
+  let url = `${config.regSearchApiUrl}/businesses/search/facets?start=0&rows=20`
+  url += `&categories=legalType:${legalType}`
+  url += `&query=value:${encodeURIComponent(queryStr)}`
+
+  const response = await $fetch<RegSearchResponse>(url, {
```
JazzarKarim, line 12: I had a discussion with Riyaz and Mihai. Here's the reason for this change:

The newer endpoint (v2), has one extra field which is the parties array:
image

At first, I was thinking that it should be OK with no changes as we're just providing this extra array with the UI not doing anything with it.

But, after speaking with Mihai, there's kind of a security/privacy issue. According to him, that proprietor name is what somebody could use to affiliate that firm's record to their account. So, the safest option is to just get rid of that in which that old endpoint does exactly that.

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-44-nvnamykh.web.app

business-registry-dashboard/app/utils/reg-search.ts
severinbeauvais: Did you check whether there's a URL parameter that would suppress the parties in the response?

If there isn't then I'm OK with this for now, but please make sure there exists a ticket to change to V2 at some point, and perhaps look up the other similar tickets (23768 and 23555) and add a comment explaining the security issue.

Ultimately, this should bubble up to the team that manages this endpoint so the security issue is handled at the source instead of forcing us to use a deprecated endpoint -- please follow up with Kial and/or create a ticket (for Melissa's team, I think).

business-registry-dashboard/app/utils/reg-search.ts
```diff
@@ -4,23 +4,16 @@ export async function regSearch (queryStr: string): Promise<RegSearchResult[]> {
   const accountStore = useConnectAccountStore()
   const token = await keycloak.getToken()
 
-  const response = await $fetch<RegSearchResponse>(`${config.regSearchApiUrl}/search/businesses`, {
-    method: 'POST',
+  const legalType = 'A,BC,BEN,C,CBEN,CC,CCC,CP,CUL,FI,GP,LL,LLC,LP,PA,S,SP,ULC,XCP,XL,XP,XS'
+  let url = `${config.regSearchApiUrl}/businesses/search/facets?start=0&rows=20`
```
severinbeauvais, line 8: I'm unable to test with the preview URL, but it looks like this is still using the v2 endpoint. The only difference I see here is the old code is a POST and the new code is a GET. You're using the same `regSearchApiUrl` in both cases.

image

severinbeauvais: PS There's probably an API key missing in the GCP config to make this work.

severinbeauvais:
What would it take to rename this repo? How about `business-registry-ui` as per the GCP config?

image

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-44-nvnamykh.web.app

business-registry-dashboard/app/utils/reg-search.ts
JazzarKarim: It's because I missed updating the vaults. My bad. Should be good now (v1 endpoint).

JazzarKarim:
> What would it take to rename this repo? How about `business-registry-ui` as per the GCP config?
> 
> image

I'm not sure how much effort it would take. I'll probably need to talk to Patrick or Thor.

business-registry-dashboard/app/utils/reg-search.ts
severinbeauvais: I'm still getting a 401, but the update vault info looks good.

JazzarKarim: The 401 error in the preview site has been there since the beginning. I tried literally everything at one point but couldn't quite get to fix it.

severinbeauvais: Add it to the list of questions for Patrick 😆 

Resolving.

business-registry-dashboard/app/utils/reg-search.ts
JazzarKarim: Sev, I'm actually closing this PR and not doing anything here. Here's why:

I spoke to Kial about this endpoint. Apparently, they are aware of this security issue and hence why the parties that are being returned, are only the public parties (proprietors / partners). This endpoint will ONLY ever return public parties which are all available to everyone. He said there are no security/privacy issues at all.

So, I'm staying on this endpoint and not doing any changes. OK for me to close this PR?

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-44-5mmpzlsq.web.app

business-registry-dashboard/app/utils/reg-search.ts
```diff
@@ -3,24 +3,18 @@ export async function regSearch (queryStr: string): Promise<RegSearchResult[]> {
   const config = useRuntimeConfig().public
   const accountStore = useConnectAccountStore()
   const token = await keycloak.getToken()
+  const ldStore = useConnectLaunchdarklyStore()
 
-  const response = await $fetch<RegSearchResponse>(`${config.regSearchApiUrl}/search/businesses`, {
-    method: 'POST',
+  const legalType = ldStore.getStoredFlag(LDFlags.AllowableBusinessSearchTypes)
```
severinbeauvais, line 8: Oh! This explains why the V2 endpoint searched for everything.

severinbeauvais (approved):

business-registry-dashboard/app/utils/reg-search.ts
JazzarKarim: Yup, found this small bug along the way. This should the FF now.

PR #56, opened 2025-01-21
feat: added support for google analytics

JazzarKarim:
https://github.com/bcgov/entity/issues/25401

Added support for google analytics

@pwei1018 Is this enough to assume that this change works?
image


JazzarKarim (commented):
One sec, fixing those tests...

JazzarKarim:
> One sec, fixing those tests...

Tests fixed 👍

pwei1018 (approved):

pwei1018:
Looks good to me. Thanks.

business-registry-dashboard/vitest.config.ts
```diff
@@ -24,7 +24,17 @@ export default defineVitestConfig({
     environmentOptions: {
       nuxt: {
         rootDir: fileURLToPath(new URL('./', import.meta.url)),
-        domEnvironment: 'happy-dom'
+        domEnvironment: 'happy-dom',
```
severinbeauvais, line 27: Why is this different than what you did in the other PR?

business-registry-dashboard/devops/vaults.env
```diff
@@ -28,3 +28,8 @@ NUXT_REGISTRIES_SEARCH_API_URL="op://API/$APP_ENV/registries-search-api/REGISTRI
 NUXT_REGISTRIES_SEARCH_API_VERSION="op://API/$APP_ENV/registries-search-api/REGISTRIES_SEARCH_API_VERSION"
 NUXT_REGISTRIES_SEARCH_API_KEY="op://API/$APP_ENV/registries-search-api/REGISTRIES_SEARCH_API_KEY"
 NUXT_X_API_KEY="op://web-url/$APP_ENV/business-registry-ui/X_API_KEY"
+
+#google analytics
+NUXT_GTAG_ID="op://google-analytics/$APP_ENV/developer/GTAG"
+NUXT_GTAG_DEBUG="op://google-analytics/$APP_ENV/developer/GTAG_DEBUG"
+NUXT_GTM_ID="op://google-analytics/$APP_ENV/developer/GTM"
```
severinbeauvais, line 35: Hey @pwei1018 , can you give me permissions to this vault in 1Password?

severinbeauvais: Should different UIs use different GA ids?

pwei1018: no, only 1 GA ids.

business-registry-dashboard/vitest.config.ts
JazzarKarim: It's the exact same change! How different do you mean?

severinbeauvais: In the other PR, it looks like this:

image

Granted, you didn't change this part. Should it be the same between projects?

JazzarKarim: Ahhh, I see.

I guess what the other project is doing is that it uses whatever is set in the environment variable VITEST_DOM_ENV (which must be either 'happy-dom' or 'jsdom'), and if that's not set, default to 'happy-dom'.

This here just goes with happy-dom always. In a way both make sense and both would work. Do you want me to change it here to stay consistent?

severinbeauvais: Consistency is good, but so is simplicity. Do we even need to set a different DOM Environment???

JazzarKarim: I don't know, I can ask Kial as I can see it was his code there. I would say no though. We probably don't need a different DOM environment.

severinbeauvais: Ask around a bit. Maybe it's good to have this option. If not, leave as-is.

PR #60, opened 2025-01-29
fixed: incorp now link to show dialog error if NR not processed yet

JazzarKarim:
https://github.com/bcgov/entity/issues/25427

Before I approved my NR:
before approval nr

If you take a look at the console, the expiration date is null. As long as that expiration date is null, the NR isn't fully ready to be consumed yet. So, we'll show that nice error dialog.

### DIFF
--- a/business-registry-dashboard/app/pages/incorporateNow.vue
+++ b/business-registry-dashboard/app/pages/incorporateNow.vue
@@ -81,7 +81,7 @@ onMounted(async () => {
     // Check basic eligibility conditions
     const isEligibleRequest = business.nameRequest.requestActionCd === NrRequestActionCodes.NEW_BUSINESS &&
       ELIGIBLE_CORP_TYPES.includes(business.nameRequest.legalType) &&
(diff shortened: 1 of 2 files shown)


business-registry-dashboard/app/pages/incorporateNow.vue
```diff
@@ -80,7 +80,8 @@ onMounted(async () => {
 
     // Check basic eligibility conditions
     const isEligibleRequest = business.nameRequest.requestActionCd === NrRequestActionCodes.NEW_BUSINESS &&
-      ELIGIBLE_CORP_TYPES.includes(business.nameRequest.legalType)
+      ELIGIBLE_CORP_TYPES.includes(business.nameRequest.legalType) &&
+      nameRequest.expirationDate
```
severinbeauvais, line 84: I'd be tempted to use `!!` on this property to convert it to a boolean.

severinbeauvais (approved):

PR #70, opened 2025-02-18
22683 remove delete continuation functionality

TVWerdal:
Issue #: https://github.com/bcgov/entity/issues/22683

Description of changes:

- Updated logic so that Continue In's would be successfully deleted or removed (have affiliations removed). 
- Added in wording in accordance with the Figma design here: https://www.figma.com/design/rNe2jn8wyFykKRzwiBo6ar/Continuation-In-V6?node-id=2188-39977&p=f&t=e3LnVAu9pjqKCmAb-0
- Refactored some related code

Also addresses following tickets:
Issue #: https://github.com/bcgov/entity/issues/24079
Issue #: https://github.com/bcgov/entity/issues/24986

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-filings-ui license (Apache 2.0).

### DIFF
--- a/business-registry-dashboard/app/components/Modal/RemoveBusiness/Generic.vue
+++ b/business-registry-dashboard/app/components/Modal/RemoveBusiness/Generic.vue
@@ -9,6 +9,13 @@ defineEmits<{
   confirm: [void]
   close: [void]
 }>()
+function selectDescription (removeBusinessPayload: RemoveBusinessPayload) {
+  // Different discriptions for Continuation In's if they have a name request
+  if (removeBusinessPayload.business.corpType.code === CorpTypes.CONTINUATION_IN && removeBusinessPayload.business.nameRequest) {
+    return `modal.removeBusiness.generic.${removeBusinessPayload.business.corpType.code}.descriptionNamed`
+  }
+  return `modal.removeBusiness.generic.${removeBusinessPayload.business.corpType.code}.description`
+}
 </script>
 <template>
   <div class="flex flex-col items-center gap-4 text-center">
@@ -16,7 +23,7 @@ defineEmits<{
     <h2 class="text-xl font-semibold">
       {{ $t(`modal.removeBusiness.generic.${removeBusinessPayload.business.corpType.code}.title`) }}
     </h2>
-    <p>{{ $t(`modal.removeBusiness.generic.${removeBusinessPayload.business.corpType.code}.description`) }}</p>
+    <p>{{ $t(selectDescription(removeBusinessPayload)) }}</p>
     <div class="mt-2 flex flex-wrap items-center justify-center gap-4">
       <UButton
         :block="isSmallScreen"
--- a/business-registry-dashboard/app/composables/useBrdModals.ts
+++ b/business-registry-dashboard/app/composables/useBrdModals.ts
@@ -65,15 +65,10 @@ export const useBrdModals = () => {
 
   function openBusinessRemovalConfirmation (removeBusinessPayload: RemoveBusinessPayload) {
     const type = removeBusinessPayload.business.corpType.code
-    if ([CorpTypes.NAME_REQUEST,
-      CorpTypes.INCORPORATION_APPLICATION,
-      CorpTypes.AMALGAMATION_APPLICATION,
-      CorpTypes.REGISTRATION,
-      CorpTypes.PARTNERSHIP,
-      CorpTypes.SOLE_PROP].includes(type)) {
-      modal.open(ModalRemoveBusiness, { removeBusinessPayload, type: 'generic' })
-    } else {
+    if (type === CorpTypes.COOP) {
       modal.open(ModalRemoveBusiness, { removeBusinessPayload, type: 'passcode' })
+    } else {
+      modal.open(ModalRemoveBusiness, { removeBusinessPayload, type: 'generic' })
     }
   }
 
--- a/business-registry-dashboard/app/stores/affiliations.ts
+++ b/business-registry-dashboard/app/stores/affiliations.ts
@@ -75,7 +75,10 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () => 
   /* Check if Business can be deleted safely (i.e. does not have a review record) */
   function canBusinessBeDeleted (payload: RemoveBusinessPayload) {
     // For now only including Continuation Ins where only draft records can be deleted
-    if (payload.business.corpType.code === CorpTypes.CONTINUATION_IN && payload.business?.draftStatus !== EntityStates.DRAFT) {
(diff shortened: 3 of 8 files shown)


business-registry-dashboard/app/composables/useBrdModals.ts
TVWerdal, line 1: Talking with Janice and Jacqueline, they noted that the password option SHOULD NOT show up for anything but coop, so I simplified this function.

TVWerdal:
Eh, hold on reviewing this - just noticed one of the unit tests broke. I'll fix that up.

TVWerdal:
> Eh, hold on reviewing this - just noticed one of the unit tests broke. I'll fix that up.

Fixed the Unit test, forgot to update this when the passcode model was no longer the default.

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-70-gkgz3g66.web.app

business-registry-dashboard/app/components/Modal/RemoveBusiness/Generic.vue
```diff
@@ -9,14 +9,21 @@ defineEmits<{
   confirm: [void]
   close: [void]
 }>()
+function selectDescription (removeBusinessPayload: RemoveBusinessPayload) {
+  // Different discriptions for Continuation In's if they have a name request
+  if (removeBusinessPayload.business.corpType.code === CorpTypes.CONTINUATION_IN && removeBusinessPayload.business.nameRequest) {
```
severinbeauvais, line 14: What about the other 3 continuation in types (CBEN, CCC, CUL)?

business-registry-dashboard/app/stores/affiliations.ts
```diff
@@ -69,6 +72,16 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () =>
     }
   }
 
+  /* Check if Business can be deleted safely (i.e. does not have a review record) */
+  function canBusinessBeDeleted (payload: RemoveBusinessPayload) {
+    // For now only including Continuation Ins where only draft records can be deleted
+    if (payload.business.corpType.code === CorpTypes.CONTINUATION_IN && payload.business?.draftStatus !== EntityStates.DRAFT) {
```
severinbeauvais, line 78: Note that the second condition will automatically be True if business is falsy (ie, undefined !== DRAFT). Ditto with draftStatus.

I get around this by first ensuring `payload.business?.draftStatus` is truthy.

business-registry-dashboard/app/components/Modal/RemoveBusiness/Generic.vue
severinbeauvais: Oh, never mind. This is one of those overloaded values.

image

Resolving.

business-registry-dashboard/tests/unit/composables/useBrdModals.test.ts
```diff
@@ -162,7 +162,10 @@ describe('useBrdModals', () => {
         CorpTypes.AMALGAMATION_APPLICATION,
         CorpTypes.REGISTRATION,
         CorpTypes.PARTNERSHIP,
-        CorpTypes.SOLE_PROP].forEach((item) => {
+        CorpTypes.SOLE_PROP,
+        CorpTypes.CONTINUE_IN,
```
severinbeauvais, line 166: This seems weird. You have some corp types in this list and some overloaded types.

Why are Partnership and Sole Prop in here?

Should line 166 actually be CONTINUATION_IN?

business-registry-dashboard/app/stores/affiliations.ts
```diff
@@ -49,13 +50,15 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () =>
...
+      CorpTypes.REGISTRATION,
+      CorpTypes.CONTINUATION_IN
     ].includes(payload.business.corpType.code)) {
       const filingResponse = await getFilings(payload.business.businessIdentifier)
       if (filingResponse) {
         const filingId = filingResponse.filing?.header?.filingId // TODO: fix typing
         // If there is a filing delete it which will delete the affiliation, else delete the affiliation
-        if (filingId) {
+        const deleteBusiness = canBusinessBeDeleted(payload)
+        if (filingId && deleteBusiness) {
```
JazzarKarim, line 61: So in here, we only enter this block if it's a continuation in? Do we want to do that? What about other filing types?

severinbeauvais: Are you responding to something?

To answer you: this block is for bootstrap filings only. I guess they don't need to be unaffiliated because the NR will be show instead.

But, `canBusinessBeDeleted` needs to be True for more than just draft continuation ins.

TVWerdal: canBusinessBeDeleted is always true unless it is a CTMP which is not in DRAFT state. Originally it was if there was a filingId it could be deleted. That should still be true for everything except a CTMP that has progressed further than DRAFT state. That being said, I don't know if there are other cases for other filings we would want to catch here and redirect to removing affiliations rather than deleting. If there are any known cases I can add them.

severinbeauvais: I can't think of any other items that we'd want to unaffiliate instead of delete.

business-registry-dashboard/tests/unit/composables/useBrdModals.test.ts
TVWerdal: I don't think these values matter as much with the new functionality. Originally there was a strict list of CorpTypes that would get the "generic" remove modal instead of the default "passcode" remove modal. Jacqueline and Janis noted that the passcode model should only ever show for COOPs. So I swapped that around, now every case will use the removal module (wording changed for delete or remove) and COOP is the only special case that prompts the passcode model. For testing purposes, this ensures that everything in the list, and even future CorpTypes receive the 'generic' remove modal. COOP is the special case that has its own test following this one. Does that make sense? Or do we want to re-evaluate/change the test?

severinbeauvais: I'd prefer to see the tests validate apples and apples, not apples and oranges. I think it's worth no more than a few hours to clean this up. Up to you.

JazzarKarim (approved):
Aside from Sev's comment, looks good 👍

business-registry-dashboard/tests/unit/composables/useBrdModals.test.ts
severinbeauvais: Thayne and I discussed this in Teams.

I'll approve this PR and leave this item with Thayne to resolve as he best sees fit.

severinbeauvais (approved):

meawong (approved):

PR #71, opened 2025-02-24
feat: redirect users with non-active accounts to account info page

JazzarKarim:
https://github.com/bcgov/entity/issues/25956

Now if a user doesn't have an active account, it will be redirected to the account info page.

### DIFF
--- a/business-registry-dashboard/app/layouts/dashboard.vue
+++ b/business-registry-dashboard/app/layouts/dashboard.vue
@@ -42,9 +42,13 @@ watch(() => route.params.orgId, (orgId) => {
 onMounted(() => {
   // Redirect unauthenticated users to login page with current URL as redirect target
   if (!isAuthenticated.value) {
-    const registryHomeURL = useRuntimeConfig().public.registryHomeURL
+    const registryHomeURL = config.registryHomeURL
     const redirectUrl = encodeURIComponent(window.location.href)
     window.location.href = `${registryHomeURL}/login/?return=${redirectUrl}`
+  }
+  if (accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
+    const accountId = accountStore.currentAccount.id
+    window.location.href = `${config.authWebUrl}/account/${accountId}/settings/account-info`
   }
 })
 </script>
--- a/business-registry-dashboard/package.json
+++ b/business-registry-dashboard/package.json
@@ -2,7 +2,7 @@
   "name": "business-registry-dashboard",
   "private": true,
   "type": "module",
-  "version": "1.0.16",
+  "version": "1.0.17",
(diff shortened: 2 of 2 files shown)


bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-71-apzcrfi4.web.app

business-registry-dashboard/app/layouts/dashboard.vue
```diff
@@ -42,10 +42,14 @@ watch(() => route.params.orgId, (orgId) => {
...
   // Redirect unauthenticated users to login page with current URL as redirect target
   if (!isAuthenticated.value) {
-    const registryHomeURL = useRuntimeConfig().public.registryHomeURL
+    const registryHomeURL = config.registryHomeURL
     const redirectUrl = encodeURIComponent(window.location.href)
     window.location.href = `${registryHomeURL}/login/?return=${redirectUrl}`
   }
+  if (accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
+    const accountId = accountStore.currentAccount.id
+    window.location.href = `${config.authWebUrl}/account/${accountId}/settings/account-info`
```
severinbeauvais, line 51: Is there a central place that these "destinations" could be set? I'm seeing duplication in our other Nuxt project and thinking that having all these in 1 place would be nice.

No need to change this here or now. Keep it in the back of your mind for later.

severinbeauvais (approved):

business-registry-dashboard/app/layouts/dashboard.vue
severinbeauvais: PS How about a code comment on this block?

JazzarKarim: > Is there a central place that these "destinations" could be set? I'm seeing duplication in our other Nuxt project and thinking that having all these in 1 place would be nice.
> 
> No need to change this here or now. Keep it in the back of your mind for later.

Yup, this makes sense. One idea would be to create a URL service that handles URL construction. Or maybe we can create an External Routes config file. Something like:
```typescript
export const EXTERNAL_ROUTES = {
  AUTH: {
    ACCOUNT_SETTINGS: (accountId: string) => `/account/${accountId}/settings/account-info`,
    xxx: '/xxx',
    yyy: '/yyy'
}
```

Then we would
```typescript
window.location.href = `${config.authWebUrl}${EXTERNAL_ROUTES.AUTH.ACCOUNT_SETTINGS(accountId)}`
```

We can for sure do something about that in the other repo and here!

TVWerdal (approved):
Looks good

ArwenQin (approved):
LGTM

meawong (approved):

PR #73, opened 2025-02-27
feat: name request expired in BRD (tooltip/icon) + misc fixes

JazzarKarim:
https://github.com/bcgov/entity/issues/26194

Screenshot (229)
inval cont in app

Screenshot (230)

### DIFF
--- a/business-registry-dashboard/app/components/Table/AffiliatedEntity/Action.vue
+++ b/business-registry-dashboard/app/components/Table/AffiliatedEntity/Action.vue
@@ -307,9 +307,11 @@ const handleApprovedNameRequest = (item: Business, nrRequestActionCd?: NrRequest
 
 async function redirect (item: Business) {
   if (isTemporaryBusiness(item)) { // handle if temp business
-    if (isExpired(item, CorpTypes.CONTINUATION_IN)) {
-      // Show error dialog for continuation application with expired name request
-      brdModal.openInvalidContinuationApplication()
+    // Check for expired name requests
+    if (isExpired(item, item.corpType.code === CorpTypes.CONTINUATION_IN ? CorpTypes.CONTINUATION_IN : undefined)) {
+      // Use tempDescription to get the full application type name
+      const applicationType = tempDescription(item)
+      brdModal.openInvalidFilingApplication(applicationType)
       return
     }
     affNav.goToDashboard(item.businessIdentifier)
--- a/business-registry-dashboard/app/components/Table/AffiliatedEntity/Action.vue
+++ b/business-registry-dashboard/app/components/Table/AffiliatedEntity/Action.vue
@@ -11,6 +11,7 @@ const accountStore = useConnectAccountStore()
 const affStore = useAffiliationsStore()
 const { t } = useI18n()
 const ldStore = useConnectLaunchdarklyStore()
+const brdModal = useBrdModals()
 
 const props = defineProps<{
   affiliations: Business[],
@@ -306,6 +307,11 @@ const handleApprovedNameRequest = (item: Business, nrRequestActionCd?: NrRequest
 
 async function redirect (item: Business) {
   if (isTemporaryBusiness(item)) { // handle if temp business
+    if (isExpired(item, CorpTypes.CONTINUATION_IN)) {
+      // Show error dialog for continuation application with expired name request
+      brdModal.openInvalidContinuationApplication()
+      return
+    }
     affNav.goToDashboard(item.businessIdentifier)
   } else if (isNameRequest(item)) { // handle if name request
     if (affiliationStatus(item) === NrDisplayStates.APPROVED) {
--- a/business-registry-dashboard/app/layouts/dashboard.vue
+++ b/business-registry-dashboard/app/layouts/dashboard.vue
@@ -51,7 +51,7 @@ onMounted(() => {
   // We check for valid IDs (>= 0) to ensure we only redirect accounts that exist in the system
   if (accountStore.currentAccount &&
       +accountStore.currentAccount.id >= 0 &&
-      accountStore.currentAccount.accountStatus === AccountStatus.ACTIVE) {
+      accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
     const accountId = accountStore.currentAccount.id
     window.location.href = `${config.authWebUrl}/account/${accountId}/settings/account-info`
   }
--- a/business-registry-dashboard/app/layouts/dashboard.vue
+++ b/business-registry-dashboard/app/layouts/dashboard.vue
@@ -48,7 +48,9 @@ onMounted(() => {
   }
   // Redirect inactive accounts to the account settings page
   // This ensures users can't access the dashboard until their account is activated
-  if (accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
+  if (accountStore.currentAccount &&
+      accountStore.currentAccount.id &&
+      accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
     const accountId = accountStore.currentAccount.id
     window.location.href = `${config.authWebUrl}/account/${accountId}/settings/account-info`
   }
--- a/business-registry-dashboard/app/layouts/dashboard.vue
+++ b/business-registry-dashboard/app/layouts/dashboard.vue
@@ -48,7 +48,9 @@ onMounted(() => {
   }
   // Redirect inactive accounts to the account settings page
   // This ensures users can't access the dashboard until their account is activated
-  if (accountStore.currentAccount?.id &&
+  // We check for valid IDs (>= 0) to ensure we only redirect accounts that exist in the system
+  if (accountStore.currentAccount &&
+      Number(accountStore.currentAccount.id) >= 0 &&
       accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
     const accountId = accountStore.currentAccount.id
     window.location.href = `${config.authWebUrl}/account/${accountId}/settings/account-info`
--- a/business-registry-dashboard/app/layouts/dashboard.vue
+++ b/business-registry-dashboard/app/layouts/dashboard.vue
@@ -48,8 +48,7 @@ onMounted(() => {
   }
   // Redirect inactive accounts to the account settings page
   // This ensures users can't access the dashboard until their account is activated
-  if (accountStore.currentAccount &&
-      accountStore.currentAccount.id &&
+  if (accountStore.currentAccount?.id &&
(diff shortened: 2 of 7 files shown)


business-registry-dashboard/app/layouts/dashboard.vue
```diff
@@ -48,7 +48,9 @@ onMounted(() => {
   }
   // Redirect inactive accounts to the account settings page
   // This ensures users can't access the dashboard until their account is activated
-  if (accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
+  if (accountStore.currentAccount &&
+      accountStore.currentAccount.id &&
+      accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
```
JazzarKarim, line 53: This was a bug with a previous PR of mine. What was happening is on INITIAL sign in only, the currentAccount has not been defined yet. So what was happening is, we were always going to the account info page with something like `https://dev.account.bcregistry.gov.bc.ca/account/undefined/settings/account-info`

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-73-0p7xpi1h.web.app

business-registry-dashboard/app/utils/affiliations.ts
```diff
@@ -169,14 +169,32 @@ export const affiliationStatus = (business: Business): string => {
...
-    return isDraft(affiliationStatus(item)) && (item.nameRequest && (item.nameRequest.expirationDate !== null) &&
-    (new Date(item.nameRequest.expirationDate) < new Date())) && isIA(affiliationType(item))
-  } else {
+/**
+ * Checks if an item has an expired Name Request
+ * @param item - The business to check
+ * @param type - Optional entity type to check for
+ * @returns True if the item has an expired Name Request matching the criteria
+ */
+export const isExpired = (item: Business, type?: CorpTypes): boolean => {
```
JazzarKarim, line 178: I took the liberty to refactor this function. I also noticed that the alert icon was not working for amalgamations. This now will fix it across the board for all filing types.

business-registry-dashboard/app/components/Table/AffiliatedEntity/Action.vue
```diff
@@ -306,6 +307,11 @@ const handleApprovedNameRequest = (item: Business, nrRequestActionCd?: NrRequest
 
 async function redirect (item: Business) {
   if (isTemporaryBusiness(item)) { // handle if temp business
+    if (isExpired(item, CorpTypes.CONTINUATION_IN)) {
+      // Show error dialog for continuation application with expired name request
+      brdModal.openInvalidContinuationApplication()
+      return
+    }
```
severinbeauvais, line 314: Why isn't this common code for IAs / Registrations / Amalgamations / Continuation Ins? All can suffer the same fate.

business-registry-dashboard/app/layouts/dashboard.vue
severinbeauvais: You could combine the first and second line (with optional chaining).

The catch (which you avoided) is that the left side could be undefined, which will not equal to the right side, thus being a true condition and executing this code block.

business-registry-dashboard/app/utils/affiliations.ts
```diff
@@ -169,14 +169,32 @@ export const affiliationStatus = (business: Business): string => {
...
+  // Return false if there's no expiration date
+  if (!item.nameRequest?.expirationDate) {
+    return false
+  }
+
+  // Check if the expiration date is in the past
+  const isExpiredDate = new Date(item.nameRequest.expirationDate) < new Date()
```
severinbeauvais, line 185: I cringe when I see this because `new Date()` creates a date object according to the user's browser, which uses the user's operating system clock, which could be wrong or spoofed.

But this issue is in a lot of places in this UI, so I think it needs another ticket to fix.

business-registry-dashboard/app/components/Table/AffiliatedEntity/Action.vue
severinbeauvais: Should this code be in getDetails() instead?

JazzarKarim: > Why isn't this common code for IAs / Registrations / Amalgamations / Continuation Ins? All can suffer the same fate.

Confirming with Jacqueline on this. I think this is exclusive to Continuation Ins but I could be wrong.

> Should this code be in getDetails() instead?

No. This is part of when the action button (Resume Application) is clicked. The `getDetails()` function is a part of the status column (status, icon, tooltip).

JazzarKarim: > Why isn't this common code for IAs / Registrations / Amalgamations / Continuation Ins? All can suffer the same fate.

> Confirming with Jacqueline on this. I think this is exclusive to Continuation Ins but I could be wrong.


Spoke to Jacqueline, she said this would be great if we can have it. Updated the code. Now:
image
image
image

business-registry-dashboard/app/utils/affiliations.ts
JazzarKarim: I totally agree with you. I fixed this here. Let me know what you think.

We have moment-timezone in this project so I might as well utilize that to the max. My solution should be robust and prevent any sort of manipulation from the user.

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-73-0p7xpi1h.web.app

TVWerdal (approved):
Nice work extending this to more filing types. Looks good to me.

business-registry-dashboard/app/layouts/dashboard.vue
severinbeauvais: Looks good.

business-registry-dashboard/app/layouts/dashboard.vue
```diff
@@ -48,8 +48,7 @@ onMounted(() => {
   }
   // Redirect inactive accounts to the account settings page
   // This ensures users can't access the dashboard until their account is activated
-  if (accountStore.currentAccount &&
-      accountStore.currentAccount.id &&
+  if (accountStore.currentAccount?.id &&
       accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
```
severinbeauvais, line 52: Doh! I guess the previous code was better because it validated that currentAccount was truthy (needed for accountStatus).

What's the logic here -- does id have to be truthy as well as account status not active?

(PS I don't think we have id=0 but if we did, that would be falsy.)

business-registry-dashboard/app/utils/affiliations.ts
severinbeauvais: Nice code but Moment also uses the browser's clock.

The solution is to use a known good clock. In other UIs, I query the webserver for the time, but that's clunky. I haven't seen a better solution.

One approach is to keep using the browser's clock as long as the code is safe (won't allow the user to do anything they shouldn't). I think the current code is safe; with a modified system clock, the user could bypass this check, but Business Dashboard UI will also check for an expired NR (though I think it, too, uses the system clock), and ultimately Legal API should not allow use of an expired NR (I think we can trust Legal API's clock).

business-registry-dashboard/app/layouts/dashboard.vue
JazzarKarim: > What's the logic here -- does id have to be truthy as well as account status not active?

Logic is:
Check first if `currentAccount` exists. After that, check if id is truthy (needed in the redirection). If both are OK, then check finally that status. ID has to be truthy as well as account status not active for this to work. If ID is undefined, the redirection will be wrong. I'm thinking I go back to the original code I had?

severinbeauvais: How about this?
```
  if (accountStore.currentAccount &&
      accountStore.currentAccount.id >= 0 &&
      accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
```

Shorter code might work but this is explicit and therefore clearer to me. And/or add comments :)

business-registry-dashboard/app/utils/affiliations.ts
JazzarKarim: > The solution is to use a known good clock. In other UIs, I query the webserver for the time, but that's clunky. I haven't seen a better solution.

I see.

I never given this scenario much thought actually. Good to know, thanks for the explanation Sev. You're right about the Business Dashboard UI checking for the expired NR as well. It probably uses a similar process as here. Legal API will be the last check in case something is really up. Got it.

business-registry-dashboard/app/layouts/dashboard.vue
severinbeauvais: image

JazzarKarim: > image

Beautiful. Makes sense to me!

JazzarKarim: Tested the code as well, works well.

business-registry-dashboard/app/utils/affiliations.ts
TVWerdal: I didn't know this either, but it makes sense as it is JS. I'm just surprised there isn't a common plugin that would just ping a configurable service.

business-registry-dashboard/app/layouts/dashboard.vue
severinbeauvais: `id` isn't already a number?

You could use unary operator `+` 😄

JazzarKarim: Sweet! Will keep this in mind moving forward for cleaner code 🧙

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-73-0p7xpi1h.web.app

business-registry-dashboard/app/utils/affiliations.ts
severinbeauvais: I haven't researched this in a couple of years. Maybe today there's a npm library that provides a known good clock without being async (ie, without pinging something in real time).

https://www.google.com/search?q=ntp+js

severinbeauvais: I would leave the original code for now and look for a general solution separately from this ticket.

JazzarKarim: OK, I'll create a techdebt ticket for this.

JazzarKarim: New ticket: https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/26245

severinbeauvais (approved):

PR #75, opened 2025-03-03
fix: fix filters clearing if window is shrunk and vice versa

JazzarKarim:
https://github.com/bcgov/entity/issues/26186

After pinpointing where the issue was, it was just deleting that line.

### DIFF
--- a/business-registry-dashboard/app/stores/affiliations.ts
+++ b/business-registry-dashboard/app/stores/affiliations.ts
@@ -242,7 +242,6 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () => 
   watchDebounced(
     width,
     (newVal) => {
-      resetFilters() // reset filters so active filters do not get hidden when screen size changes
       if (newVal < 640) {
         // Mobile view
         visibleColumns.value = [nameColumn, actionColumn]
--- a/business-registry-dashboard/package.json
+++ b/business-registry-dashboard/package.json
@@ -2,7 +2,7 @@
   "name": "business-registry-dashboard",
   "private": true,
   "type": "module",
-  "version": "1.0.22",
+  "version": "1.0.21",
   "scripts": {
     "build-check": "nuxt build",
     "build": "nuxt generate",
--- a/business-registry-dashboard/package.json
+++ b/business-registry-dashboard/package.json
@@ -2,7 +2,7 @@
   "name": "business-registry-dashboard",
   "private": true,
   "type": "module",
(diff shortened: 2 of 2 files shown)


business-registry-dashboard/app/stores/affiliations.ts
```diff
@@ -242,7 +242,6 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () =>
   watchDebounced(
     width,
     (newVal) => {
-      resetFilters() // reset filters so active filters do not get hidden when screen size changes
```
JazzarKarim, line 245: Fix is straightforward. Filtered were being reset on window width change:
```typescript
  function resetFilters () {
    affiliations.filters.businessName = ''
    affiliations.filters.businessNumber = ''
    affiliations.filters.type = ''
    affiliations.filters.status = ''
  }
```

Removing this line will fix the issue.

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-75-vtob80bq.web.app

business-registry-dashboard/app/stores/affiliations.ts
JazzarKarim: Also, this was causing the BRD to slow down on every slight change in the window's size. I think it's much better user experience now.

business-registry-dashboard/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "business-registry-dashboard",
   "private": true,
   "type": "module",
-  "version": "1.0.20",
+  "version": "1.0.22",
```
severinbeauvais, line 5: What happened to 21?

business-registry-dashboard/app/stores/affiliations.ts
JazzarKarim: It is still valid. I just figured it's OK though. Why would a user apply a filter and decide to shrink the window while this filter is applied? It'll look something like this:
image

I applied the continuation application filter and shrunk the window. The filter is still there but doesn't show. In my head I think that's OK. Thoughts?

TVWerdal (approved):
Good catch, looks good to me.

ArwenQin (approved):

business-registry-dashboard/app/stores/affiliations.ts
severinbeauvais: I think that's OK, especially since the Clear Filters button is visible.

severinbeauvais (approved):

PR #77, opened 2025-03-05
feat: magic link routes for Cont In and Amalg NRs + refactor

JazzarKarim:
https://github.com/bcgov/entity/issues/26247

- Added magic link support for continuation In NRs
- Added magic link support for amalgamation NRs
- Refactored the code to make adding magic link support easier (moved common code to composable)

I added now to all routes to stay consistent. Let me know what you think.

In case we need to add any support for other filings in the future, it'll now be very easy 😁

### DIFF
--- a/business-registry-dashboard/app/composables/useMagicLinkProcessing.ts
+++ b/business-registry-dashboard/app/composables/useMagicLinkProcessing.ts
@@ -1,0 +1,141 @@
+import {
+  NrRequestActionCodes,
+  FilingTypes
+} from '@bcrs-shared-components/enums'
+
+export interface MagicLinkConfig {
+  filingType: FilingTypes
+  eligibleCorpTypes: string[]
+  requestActionCode: NrRequestActionCodes
+  errorModalTitle: string
+  errorModalDescription: string
+  errorModalDescription2: string
+}
+
+// Define a type for the response from createNamedBusiness, this fixes the typescript error
+interface BusinessResponse {
+  filing: {
+    business: {
+      identifier: string
+    }
+  }
+}
+
+/**
+ * Composable for handling magic link processing for business creation flows
+ */
+export function useMagicLinkProcessing () {
+  const affNav = useAffiliationNavigation()
+  const affStore = useAffiliationsStore()
+  const brdModal = useBrdModals()
+  const isLoading = ref(true)
+  const route = useRoute()
+
+  /**
+   * Creates a new business from a Name Request
+   * @param business - Business object containing corpType and nameRequest
+   * @param filingType - The type of filing to create
+   * @returns Promise<string> - The identifier of the newly created business
+   */
+  const createBusinessFromNR = async (business: any, filingType: FilingTypes): Promise<string> => {
+    const payload = {
+      filingType,
+      business
+    }
+
+    const filingResponse = await createNamedBusiness(payload) as BusinessResponse
+    return filingResponse.filing.business.identifier
+  }
+
+  /**
+   * Process the magic link flow:
+   * 1. Extract and validate URL parameters (nr, email, phone)
+   * 2. Fetch name request details
+   * 3. Create affiliation if not already affiliated
+   * 4. Verify business eligibility
+   * 5. Create business and redirect to dashboard if needed
+   * @param config - Configuration for the magic link processing
+   */
+  const processMagicLink = async (config: MagicLinkConfig) => {
+    try {
+      isLoading.value = true
+      // Extract query parameters from the URL
+      const { nr, email, phone } = route.query as { nr: string, email: string, phone: string }
+
+      // Ensure either email or phone is provided as contact information
+      if (!email && !phone) {
+        throw new Error('Missing contact information')
+      }
+
+      // Fetch and validate name request details
+      const nameRequest = await fetchNameRequest(nr, phone, email)
+
+      // Transform name request data to match expected business structure
+      const business = {
+        corpType: { code: nameRequest.legalType },
+        nameRequest: {
+          ...nameRequest,
+          nrNumber: nameRequest.nrNum, // Transform nrNum to nrNumber
+          requestActionCd: nameRequest.request_action_cd // Transform request_action_cd to requestActionCd
+        }
+      }
+
+      // Before the affiliation checks, add a guard clause
+      if (!business.nameRequest?.nrNumber) {
+        throw new Error('Name request number is required')
+      }
+
+      // Now TypeScript knows nrNumber is defined for all following uses
+      const nrNumber = business.nameRequest.nrNumber
+
+      // Check basic eligibility conditions
+      const isEligibleRequest = business.nameRequest.requestActionCd === config.requestActionCode &&
+        config.eligibleCorpTypes.includes(business.nameRequest.legalType) &&
+        !!nameRequest.expirationDate
+
+      if (!isEligibleRequest) {
+        throw new Error('Invalid name request')
+      }
+
+      // Load and check if the NR is already affiliated
+      await affStore.loadAffiliations()
+      const isAffiliated = isNameRequestAffiliated(affStore.affiliations.results, nrNumber)
+
+      // Create affiliation if it doesn't exist
+      if (!isAffiliated) {
+        await affStore.createNRAffiliation({
+          businessIdentifier: nrNumber,
+          ...(phone && { phone }),
+          ...(email && { email })
+        })
+      }
+
+      // Get the temporary business identifier if it exists
+      const tempBusinessIdentifier = getTempBusinessIdentifierOfNameRequest(affStore.affiliations.results, nrNumber)
+
+      // If the NR is already affiliated and a draft business exists, navigate to the dashboard with the temporary business identifier
+      if (tempBusinessIdentifier) {
+        await affNav.goToDashboard(tempBusinessIdentifier)
+      } else {
+        // Otherwise, create a new business and navigate to the dashboard with the new temporary business identifier
+        const identifier = await createBusinessFromNR(business, config.filingType)
+        await affNav.goToDashboard(identifier)
+      }
+    } catch (error) {
+      // All errors are handled with the magic link filing error modal
+      console.error(`Error in ${config.filingType} flow:`, error)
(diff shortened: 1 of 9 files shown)


bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-77-m7ouwnq2.web.app

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-77-m7ouwnq2.web.app

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-77-m7ouwnq2.web.app

business-registry-dashboard/app/composables/useMagicLinkProcessing.ts
```diff
@@ -0,0 +1,141 @@
...
+      const isEligibleRequest = business.nameRequest.requestActionCd === config.requestActionCode &&
+        config.eligibleCorpTypes.includes(business.nameRequest.legalType) &&
+        !!nameRequest.expirationDate
+
+      if (!isEligibleRequest) {
+        throw new Error('Invalid name request')
+      }
+
+      // Load and check if the NR is already affiliated
+      await affStore.loadAffiliations()
```
severinbeauvais, line 101: Would this be slow if there are many affiliations?

severinbeauvais:
> I added now to all routes to stay consistent. Let me know what you think.

I love it!

> In case we need to add any support for other filings in the future, it'll now be very easy 😁

As of today, there's just one missing. Want to add it? Firm registrations.

severinbeauvais (approved):
Looks great!

Is there still duplication from "Amalgamate Now" etc button actions?

business-registry-dashboard/app/composables/useMagicLinkProcessing.ts
JazzarKarim: Sadly yes 😞 This is the function that makes the call to the affiliations endpoint:
```typescript
      const response = await $authApi<{ entities: AffiliationResponse[] }>(
        `/orgs/${orgId}/affiliations?new=true`
      )
```

Once we implement that fix that makes it faster, this should make this faster here as well.

severinbeauvais: Can that endpoint be called with the affiliation you're looking for, to see if it exists? Something like "GET /orgs/1234/affiliations/Txxx"

Is this a second "get affiliations" call as the BRD loads up, or does it replace the other one?

TVWerdal (approved):
Much cleaner, looks good.

business-registry-dashboard/app/composables/useMagicLinkProcessing.ts
JazzarKarim: So looking at the code, this call replaces the other one. This means that we're making one call.

We need this for sure though to load the data in the BRD. If for example the magic link fails, we need the data in the BRD loaded so eventually we need to load these affiliations.

ArwenQin (approved):

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-77-m7ouwnq2.web.app

JazzarKarim:
> As of today, there's just one missing. Want to add it? Firm registrations.

Added! You can try:
https://business-registry-dev--pr-77-m7ouwnq2.web.app/registerNow?nr=NR+9340980&phone=1112223333

JazzarKarim:
> Is there still duplication from "Amalgamate Now" etc button actions?

What do you mean Sev?

severinbeauvais:
> > Is there still duplication from "Amalgamate Now" etc button actions?
> 
> What do you mean Sev?

Does the "Amalgamate Now" button still have a separate flow to create a draft filing, or could it use some of the methods in your composable? Ditto for Continue In Now and Incorporate Now.

business-registry-dashboard/app/composables/useMagicLinkProcessing.ts
severinbeauvais: It looks like there are existing endpoints to fetch specific affiliations:
- https://github.com/bcgov/sbc-auth/blob/aa0c9cd13484db63d2e03e2c5aac5f95c32658ce/auth-api/src/auth_api/resources/v1/org.py#L434
- https://github.com/bcgov/sbc-auth/blob/aa0c9cd13484db63d2e03e2c5aac5f95c32658ce/auth-api/src/auth_api/resources/v1/org.py#L452

Using one of these to look up a single affiliation would be a better design, if only because it cause the systems to do less work (and traffic).

I recommend playing with these and seeing if they work as expected are are easy to use here, and then use the regular affiliations call when loading the BRD. Thoughts?

JazzarKarim:
> Does the "Amalgamate Now" button still have a separate flow to create a draft filing, or could it use some of the methods in your composable? Ditto for Continue In Now and Incorporate Now.

Ah, I see. The button actions have different flows. They use a different function which makes an FF check to see if we support that filing type:
image

However, now that I'm looking, I think we probably don't want these Registration and IA types to be hardcoded. I'll move to use the FF probably.

severinbeauvais:
> Ah, I see. The button actions have different flows. They use a different function which makes an FF check to see if we support that filing type:

The FF check determines what action to show on the button, and after that it's the same flow, right?

> However, now that I'm looking, I think we probably don't want these Registration and IA types to be hardcoded. I'll move to use the FF probably.

Yes, no more hard-coded logic for this. We need to be flexible for Apr 1 and MVP.

business-registry-dashboard/app/composables/useMagicLinkProcessing.ts
JazzarKarim: OK, Sev, I tried doing this as I like the idea. The endpoint works and I can here instead of loading everything just load that particular affiliation using the NR number and it's faster.

I hit a dead-end though and here's why. Eventually we want to load all the affiliations. The reason is line 114:
```typescript
// Get the temporary business identifier if it exists
const tempBusinessIdentifier = getTempBusinessIdentifierOfNameRequest(affStore.affiliations.results, nrNumber)
```

We want to check if the temp business identifier is in the table. If yes, we don't want to create a new temporary business draft (or else we end up with multiple drafts if user clicks the link again and again). That endpoint will not help in this case. This function checks if the temporary business identifier is already in the table. 

I checked to see if there's an endpoint where: Given an NR number and org ID, get temp business identifier. It doesn't exist.

severinbeauvais: Yes, I understand the problem.

Look at how the Auth API code resolves the "do I show the NR or do I show the draft" issue. Check with Travis (or Kial). Maybe you could reproduce that logic here, maybe not.

I really appreciate you looking into this. It would great code if we get this working.

JazzarKarim:
> The FF check determines what action to show on the button, and after that it's the same flow, right?
Yep, more or less.

> Yes, no more hard-coded logic for this. We need to be flexible for Apr 1 and MVP.
Fixed. Now using that `ia-supported-entities-brd` FF for this.

business-registry-dashboard/app/composables/useMagicLinkProcessing.ts
JazzarKarim: OK, will try to find a solution for this.

severinbeauvais: OK, don't stay stuck for too long.

Have a look at what Namerequest UI does for the "Incorporate Now" action:
https://github.com/bcgov/namerequest/blob/main/app/src/mixins/nr-affiliation-mixin.ts

severinbeauvais: (I know this is existing code that you've simply refactored, but in addition to bad UX, I'm afraid this will fail if there are too many affiliations and we don't fix paging/filtering before MVP.)

JazzarKarim: Small note here that I found, Name request UI falls into this trap. If the user goes and clicks incorporate now again and again, the UI keeps on creating drafts. So when I go to BRD and try to delete the draft, I have to delete it 3 times. For example:

[screen-capture (24).webm](https://github.com/user-attachments/assets/730a0849-5b78-4463-8653-37c2eb69e846)

Still looking for a solution..

severinbeauvais: By the way, one of these days I'd like to change Namerequest UI to simply route to the magic link to affiliate and create drafts.

Is there an Auth API call you could use to check if a draft with the subject NR is already affiliated to the subject account?

severinbeauvais: > By the way, one of these days I'd like to change Namerequest UI to simply route to the magic link to affiliate and create drafts.

(We'll need to figure out how to create numbered bootstrap businesses.)

JazzarKarim: No, that's what I've been looking for. There isn't a call for that (I checked a lot). What I'm thinking is, I create an Auth API ticket to handle that and move this forward. Are you OK with that?

JazzarKarim: > By the way, one of these days I'd like to change Namerequest UI to simply route to the magic link to affiliate and create drafts.

This would be nice for sure. Makes sense.

severinbeauvais: > No, that's what I've been looking for. There isn't a call for that (I checked a lot). What I'm thinking is, I create an Auth API ticket to handle that and move this forward. Are you OK with that?

Yeah, let's go with this.

PS The new Auth API endpoint would be similar to `get_org_affiliation_by_business_identifier()` except for NRs instead of business ids.

JazzarKarim: > PS The new Auth API endpoint would be similar to `get_org_affiliation_by_business_identifier()` except for NRs instead of business ids.

Makes sense to me. Sounds good. 

New ticket: https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/26436

PR #114, opened 2025-04-25
27263 - Officer - Full form in edit mode

deetz99:
partial: bcgov/entity#27263

- update table actions to match v3 design
- full form in edit mode instead of sections
- remove nested forms as not needed anymore due to full form
- simplify officer store actions, add comments

full form in table edit


bcregistry-sre:
Temporary Url for review: https://business-people-dev--pr-114-cahhhgl3.web.app

thorwolpert (approved):

PR #116, opened 2025-04-30
27067 - Fixed LD targetting + wrong user reporting

JazzarKarim:
https://github.com/bcgov/entity/issues/27067

OK, here's a bit of info what I did here:

First and foremost, I moved the LD store out of the nuxt daxiom shared package. The store used to be here:
`business-ui\business-registry-dashboard\node_modules\@daxiom\nuxt-core-layer-test\app\stores`

Now, I don't know if Thor will like this but I needed to do that in order to make my own changes. The changes being correctly reporting the context to LD when an org/user is changed. 

- I added some additional `console.info` to see what's going on
- The context is updated and sent to LD on user/org change.

Here's how I tested:
I logged in as BCREG0020 which defaults to resd right. The contexts in LD are:
image
You can see up top 668 which is the org id for resd. 

Now, let's change the account to mine: 
image

The event now is passed (previously this doesn't happen):
image

Now checking in LD after waiting a bit:
image

As you can see, this org ID resembles my dev account in BCREG0020 that I just switched to.

Now finally, to test this, what I did is I enabled targeting for my org ID:
image

And it works! You can see the pagination controls below.
image
 
### NOTE: An Ad blocker prevents the app from relaying the contexts to LD. This is why we see sometimes anonymous in LD. Sev, Mihai had this issue at one point if you remember and we were wondering why. This might've been it. The ad blocker MUST be disabled to report back correctly to LD.

### DIFF
--- a/business-registry-dashboard/app/stores/connect-launchdarkly.ts
+++ b/business-registry-dashboard/app/stores/connect-launchdarkly.ts
@@ -34,16 +34,11 @@ export const useConnectLaunchdarklyStore = defineStore('brd-connect-ld-store', (
     }
   })
 
-  function updateUserContext () {
-    if (!ldClient.value) { return }
-
-    // Only update context if user is authenticated
-    if (!keycloak.isAuthenticated || !keycloak.kcUser?.keycloakGuid) {
-      console.info('Not updating LaunchDarkly context - user not authenticated')
-      return
-    }
-
+  // Create the user and org context objects for LaunchDarkly
+  function createLdContext () {
     const appName = useRuntimeConfig().public.appName
+
+    // Create user context
     const user = {
       key: keycloak.kcUser.keycloakGuid,
       firstName: keycloak.kcUser.firstName,
@@ -54,8 +49,10 @@ export const useConnectLaunchdarklyStore = defineStore('brd-connect-ld-store', (
       appSource: appName
     }
 
-    let org = { key: user.key, appSource: appName } // Default to user key if no account
+    // Default org to user key if no account
+    let org = { key: user.key, appSource: appName }
 
+    // Use account info if available
     if (accountStore.currentAccount.id) {
       org = {
         key: accountStore.currentAccount.id,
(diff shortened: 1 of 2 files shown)


bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-116-bew54xjh.web.app

business-registry-dashboard/app/stores/connect-launchdarkly.ts
```diff
@@ -0,0 +1,195 @@
...
+    isInitializing.value = true
+
+    const ldClientId = useRuntimeConfig().public.ldClientId
+    if (!ldClientId) {
+      console.info('No LaunchDarkly SDK variable set. Aborting LaunchDarkly setup.')
+      isInitializing.value = false
+      return
+    }
+
+    const appName = useRuntimeConfig().public.appName
```
severinbeauvais, line 111: This code (and below) looks like a duplicate of the code above on line 46.

business-registry-dashboard/app/stores/connect-launchdarkly.ts
```diff
@@ -0,0 +1,195 @@
...
+    ldClient,
+    ldContext,
+    ldFlagSet,
+    ldInitialized,
+    init,
+    getFeatureFlag,
+    getStoredFlag,
+    updateUserContext,
+    $reset
+  }
```
severinbeauvais, line 194: Are all of these needed outside of this module? Or could this interface be cleaned up to just 1 or 2 items?

business-registry-dashboard/app/stores/connect-launchdarkly.ts
```diff
@@ -0,0 +1,195 @@
...
+    if (accountId && ldInitialized.value) {
+      updateUserContext()
+    }
+  })
+
+  function updateUserContext () {
+    if (!ldClient.value) { return }
+
+    // Only update context if user is authenticated
+    if (!keycloak.isAuthenticated || !keycloak.kcUser?.keycloakGuid) {
```
severinbeauvais, line 41: Can this ever happen?

Does this automatically recover, or is this a real error?

business-registry-dashboard/app/stores/connect-launchdarkly.ts
```diff
@@ -0,0 +1,195 @@
...
+      // Update flag values after context change
+      ldFlagSet.value = ldClient.value?.allFlags() || {}
+      console.info('LaunchDarkly context updated successfully')
+    }).catch((error) => {
+      console.error('LaunchDarkly context update error:', error)
+    })
+  }
+
+  function init () {
+    console.info('Initializing LaunchDarkly')
```
severinbeauvais, line 84: It would be nice to minimize LD-related console output. What do you think about just 1 success/fail message? I'm not stuck on this one.

severinbeauvais:
> ### NOTE: An Ad blocker prevents the app from relaying the contexts to LD. This is why we see sometimes anonymous in LD. Sev, Mihai had this issue at one point if you remember and we were wondering why. This might've been it. The ad blocker MUST be disabled to report back correctly to LD.

I'm a bit surprised by this. Please do a bit of research on this, in case there's a workaround.

I guess using the default context will not cause any Prod issues, since only specific accounts will need targeting. I'm assuming the default LD config is fine for everyone except the exceptions.

business-registry-dashboard/app/stores/connect-launchdarkly.ts
JazzarKarim: Minimized code duplication in latest code.

business-registry-dashboard/app/stores/connect-launchdarkly.ts
JazzarKarim: Agreed, done.

business-registry-dashboard/app/stores/connect-launchdarkly.ts
JazzarKarim: No, this is theoretically won't happen. I removed.

business-registry-dashboard/app/stores/connect-launchdarkly.ts
JazzarKarim: Agreed, I put only the ones that are important.

JazzarKarim:
> > ### NOTE: An Ad blocker prevents the app from relaying the contexts to LD. This is why we see sometimes anonymous in LD. Sev, Mihai had this issue at one point if you remember and we were wondering why. This might've been it. The ad blocker MUST be disabled to report back correctly to LD.
> 
> I'm a bit surprised by this. Please do a bit of research on this, in case there's a workaround.
> 
> I guess using the default context will not cause any Prod issues, since only specific accounts will need targeting. I'm assuming the default LD config is fine for everyone except the exceptions.

I'll look into a workaround. Default context won't cause any problems as you mentioned. This will only cause problems to targeted accounts or orgs.

JazzarKarim:
@severinbeauvais Sev, if you select only the last commit, you can see the differences between the one in daxiom package and mine.

severinbeauvais (approved):

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-116-bew54xjh.web.app

bobdev-94 (approved):
LGTM

TVWerdal (approved):
Looks good

meawong (approved):

ArwenQin (approved):
Looks good!

PR #118, opened 2025-05-01
Changed LD ID to use auth one since they are similar

JazzarKarim:


business-registry-dashboard/devops/vaults.env
```diff
@@ -15,7 +15,7 @@ NUXT_BASE_URL="op://web-url/$APP_ENV/business-registry-ui/BUSINESS_REGISTRY_URL"
 NUXT_CORPORATE_ONLINE_URL="op://web-url/$APP_ENV/bcregistry/COLIN_URL"
 NUXT_BUSINESS_DASH_URL="op://web-url/$APP_ENV/business-dash/BUSINESS_DASH_URL"
 NUXT_ENVIRONMENT_HEADER="op://entity/$APP_ENV/business-registry-ui/ENVIRONMENT_HEADER"
-NUXT_LD_CLIENT_ID="op://launchdarkly/$APP_ENV/business-registry-ui/BRD_LD_CLIENT_ID"
+NUXT_LD_CLIENT_ID="op://launchdarkly/$APP_ENV/auth/AUTH_LD_CLIENT_ID"
```
severinbeauvais, line 18: These are different LD projects, right? Does the sbc-auth project have the same FFs, with the same values, as the other project?

severinbeauvais (commented):
This is weird. I'm pretty sure I was setting FFs for the BRD in the sbc-auth LD project.

business-registry-dashboard/devops/vaults.env
JazzarKarim: In 1Pass, these value are the exact same in all environments. Patrick requested that we don't have duplicate values in 1pass. Me personally, I don't mind keeping this as is. The BRD uses FFs from sbc-auth project since it was there before.

severinbeauvais (approved):
LGTM.

Keep an eye on which features are enabled/disabled when this is deployed to Test, to verify that the LD project is working correctly for you.

pwei1018 (approved):

PR #120, opened 2025-05-07
27629 - Fixed infinite loop of calls to affiliations endpoint + misc. fixes

JazzarKarim:
https://github.com/bcgov/entity/issues/27629

Fixed infinite loop of calls to affiliations endpoint when one FF is enabled

### DIFF
--- a/business-registry-dashboard/app/stores/affiliations.ts
+++ b/business-registry-dashboard/app/stores/affiliations.ts
@@ -184,8 +184,8 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () => 
   }
 
   async function loadAffiliations (): Promise<void> {
-    // Only reset if server-side filtering/pagination is disabled
-    const shouldUseServerFeatures = enableServerFiltering.value && enablePagination.value
+    // Only reset if neither server-side filtering nor pagination are enabled
+    const shouldUseServerFeatures = enableServerFiltering.value || enablePagination.value
     if (!shouldUseServerFeatures) {
       resetAffiliations()
     }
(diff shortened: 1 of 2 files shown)


bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-120-r103nytx.web.app

business-registry-dashboard/app/stores/affiliations.ts
```diff
@@ -298,7 +298,11 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () =>
       () => enablePagination.value ? affiliations.pagination : null
     ],
     async () => {
-      await loadAffiliations()
+      // Prevent infinite loop when only one feature flag is enabled
+      const isLoading = affiliations.loading
+      if (!isLoading) {
+        await loadAffiliations()
+      }
```
severinbeauvais, line 305: This is a more general fix than the FF issue, right?

This prevents a second affiliations call when one is already in progress.

However, won't this fail if I'm typing in a filter value slowly -- only the first call will succeed, and the last call (ie, the entire filter value) will be skipped?

severinbeauvais: If I'm typing in something slowly then, ideally, only the affiliations call for the last letter I type will succeed. The others would be aborted.

However, I think the API is still doing the work even if the UI cancels the call.

Check if your debounce function has any options like "leading" or "trailling", which might fix this issue.

Ref: https://lodash.com/docs/4.17.15#debounce

severinbeauvais: Tell me again which watched value is changing and triggering this function...?

severinbeauvais: As for triggering multiple concurrent affiliations calls, line 303 would prevent that, but you could also keep track that "one more" call is needed when the user stops typing or changing filters.

JazzarKarim: Please see the latest commit. So here's what I did:

- I realized we ONLY need watch debounced for 2 cases, when typing business name and business number. I limited the usage of debounce watch to only those 2.
- The `if (!affiliations.loading) {` is to prevent the API call from being made while the table is still loading. We don't want the user to search while the table is still being loaded.
- I noticed that when I filter by Active let's say, I don't see the other options anymore because we're grabbing the statuses from the filtered results:
image
I fixed that. Keep in mind the fix will work whether server side filtering is enabled or disabled. 
- Anything else, I use a normal watch because we need immediate responses. 
- Fixed infinite looping by running the watchers only when we need to. Also, don't trigger any API calls when still loading.
- Keep existing functionality of filtering locally when BOTH feature flags are off.
- When a new filter is applied, always go back to page 1. Otherwise, we might end up on a page that doesn't exist in the filtered results. This was discussed previously with Jacqueline.

I tested for all cases:
- Both FFs on. 
- Server side filtering on but pagination off
- Pagination on but server side filtering off
- Both FFs off

It should be good now for all cases.

PS: There are some issues when pagination is on but server side filtering is off. I understand that paging is not done yet? CC: @Rajandeep98

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-120-r103nytx.web.app

business-registry-dashboard/app/stores/affiliations.ts
severinbeauvais: Excellent!

business-registry-dashboard/app/stores/affiliations.ts
```diff
@@ -287,20 +287,57 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () =>
     }
   }
 
-  // Watch for changes to account, filters, or pagination with debounce
-  // This prevents multiple rapid successive API calls
-  // For example: when user is typing in the business name filter,
-  // the API is called with the final value after the user has stopped typing
-  watchDebounced(
+  // Watch for changes to account, feature flags, and pagination
```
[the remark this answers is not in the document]
JazzarKarim: Good catch. No, I didn't. I'll do that now!

business-registry-dashboard/app/stores/affiliations.ts
```diff
@@ -467,14 +504,41 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () =>
...
+  // Store all available type options from the full unfiltered results
+  const allTypeOptions = ref<string[]>([])
+
+  // Update the available options when the full results are loaded
+  watch(
+    () => affiliations.results,
+    (results) => {
+      if (results.length > 0) {
+        // Only update if we have results and no filters are applied
+        if (!hasFilters.value) {
```
severinbeauvais, line 519: Makes sense.

severinbeauvais: Thanks for the comments! Without them, I wouldn't understand why you're doing half of this 😛

severinbeauvais (approved):
Looks great! Lots of edge cases here but I think it's all good.

business-registry-dashboard/app/stores/affiliations.ts
JazzarKarim: I wouldn't understand why I did something a minute later of doing it myself if I didn't write a comment on the spot haha

PR #122, opened 2025-05-08
27629 - Fixed text filtering bug + put minimum characters

JazzarKarim:
https://github.com/bcgov/entity/issues/27629

Sev found a bug. Let's say you type NR4 and then the table starts loading. While the table is loading, if you type x after 4, the results will still be for NR4 which is very bad user experience honestly and will be probably frustrating for users. This PR fixes this. 

I also added a minimum requirement for how many characters is needed. Sev, please if you can test the bug in the temp link, that would be great. Thanks.

### DIFF
--- a/business-registry-dashboard/app/stores/affiliations.ts
+++ b/business-registry-dashboard/app/stores/affiliations.ts
@@ -32,6 +32,21 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () => 
     error: false
   })
 
+  // Track if filters changed during loading
+  const filtersChangedDuringLoading = ref(false)
+  // Store the latest filter values
+  const latestFilters = reactive({
+    businessName: '',
+    businessNumber: ''
+  })
+
+  // Helper function to check if filters have minimum required characters
+  function hasMinimumFilterCharacters (filters: { businessName: string, businessNumber: string }): boolean {
+    return filters.businessName.length >= 2 ||
+           filters.businessNumber.length >= 2 ||
+           (filters.businessName.length === 0 && filters.businessNumber.length === 0)
+  }
+
   // Pagination limit options
   const paginationLimitOptions = ref([
     { label: '50', value: 50 },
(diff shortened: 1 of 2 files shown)


severinbeauvais:
> Sev, please if you can test the bug in the temp link, that would be great. Thanks.

Looks to be functioning well. I can't reproduce it!

JazzarKarim:
> > Sev, please if you can test the bug in the temp link, that would be great. Thanks.
> 
> Looks to be functioning well. I can't reproduce it!

Awesome, thank you!

business-registry-dashboard/app/stores/affiliations.ts
```diff
@@ -32,6 +32,21 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () =>
...
+  // Store the latest filter values
+  const latestFilters = reactive({
+    businessName: '',
+    businessNumber: ''
+  })
+
+  // Helper function to check if filters have minimum required characters
+  function hasMinimumFilterCharacters (filters: { businessName: string, businessNumber: string }): boolean {
+    return filters.businessName.length >= 2 ||
+           filters.businessNumber.length >= 2 ||
```
severinbeauvais, line 46: I think a min length of 2 for business number is nice, as it allows you to search for all "BC"s.

I recommend changing min length for business name to 3 (which matches other places we do this in the code).

severinbeauvais (approved):
LGTM

business-registry-dashboard/app/stores/affiliations.ts
JazzarKarim: Yup, exactly. In my head, users will search NR or BC first right. So I thought 2 is good.

Sounds good to me about the name. I agree.

TVWerdal: Do you think there would ever be a case where someone was like "What was the name of that company again? It started with an A" and just put an A in and wonder why nothing happens? I mean I doubt there would be many people with enough businesses to run into that type of thing, but just speculating.

business-registry-dashboard/app/stores/affiliations.ts
```diff
@@ -284,6 +299,24 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () =>
       affiliations.error = true
     } finally {
       affiliations.loading = false
+
+      // Check if filters changed during loading and initiate a new search if needed
+      if (filtersChangedDuringLoading.value) {
```
TVWerdal, line 304: This works really well. I kept trying to trip it up and I couldn't. Great job.

TVWerdal (approved):

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-122-an6zahr6.web.app

business-registry-dashboard/app/stores/affiliations.ts
JazzarKarim: Awesome, appreciate you helping with the testing Thayne!

business-registry-dashboard/app/stores/affiliations.ts
JazzarKarim: Really good point. Hmmm, I think you are correct. @severinbeauvais Sev, I'm tempted to remove the minimum requirements for the name as per what Thayne just said. Thoughts?

severinbeauvais: I think tough luck for them. This isn't Google. We have the 3-char limitation in other searches already.

For a tie-breaker, ask Jacqueline or Janis.

JazzarKarim: I'm merging this PR. Both of them are away and like you mentioned, we already have that 3-char limit. It's OK.

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-122-an6zahr6.web.app

PR #128, opened 2025-05-27
27604 - Moved business-registry-dashboard into the web directory

JazzarKarim:
https://github.com/bcgov/entity/issues/27604

- Moved business-registry-dashboard from the root and into /web


severinbeauvais (approved):
I can't verify whether you missed anything, but the Checks all passed and the preview URL seems to work: 👏

thorwolpert:
thanks for moving this so quickly!
greatly appreciated

PR #129, opened 2025-05-29
28687 - Implemented the search endpoint with it's feature flag

JazzarKarim:
https://github.com/bcgov/entity/issues/28687

image

If true, use:
```typescript
let url = `/orgs/${orgId}/affiliations/search?mapping=true`
```

If false, use:
```typescript
let url = `/orgs/${orgId}/affiliations?new=true`
```

- Alphabetized LDFlags enum


bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-129-myewi05y.web.app

web/business-registry-dashboard/app/enums/ld-flags.ts
```diff
@@ -18,12 +22,9 @@ export enum LDFlags {
...
+  SentryEnable = 'sentry-enable',
   SupportedAmalgamationEntities = 'supported-amalgamation-entities',
   SupportedContinuationInEntities = 'supported-continuation-in-entities',
-  SentryEnable = 'sentry-enable',
   SupportRestorationEntities = 'supported-restoration-entities',
-  AllowableBusinessSearchTypes = 'allowable-business-search-types',
-  AllowableBusinessPasscodeTypes = 'allowable-business-passcode-types',
-  EnableAffiliationDelegation = 'enable-affiliation-delegation',
-  EnableCreditCardPremium = 'enable-credit-card-premium'
+  UseSearchEndpointToLoadAffiliations = 'use-search-endpoint-to-load-affiliations'
```
severinbeauvais, line 29: Why not use the existing FF `EnableAffiliationsServerFiltering` ?

JazzarKarim: Because if there is an issue with the new endpoint, we want to keep using the old one but keep on server side filtering. If I use the existing FF, we wouldn't be able to do that.

severinbeauvais: Does the old endpoint work correctly with server side filtering? If not then I still recommend combining these.

JazzarKarim: Yes, it works. That's what is currently happening and working.

@Rajandeep98 Can you confirm please that it should still work after your changes go through?

JazzarKarim: ok, i'll remove the new ff and use the pagination one.

JazzarKarim: Done. I'll remove that new feature flag from LaunchDarkly.

severinbeauvais: If that flag is used (enabled) anywhere,  don't remove it until after the code is deployed there.

severinbeauvais: (You can resolve this conversation if there's nothing else.)

PR #130, opened 2025-06-03
28725 - Get roles from KC token + updated role permissions

JazzarKarim:
https://github.com/bcgov/entity/issues/28725

- app version = 1.1.3
- renamed RESTORE_OR_REINSTATE -> RESTORATION_REINSTATEMENT_FILING
- replaced Authorizations call with KC roles
- deleted obsolete Auth Services call
- finalized authorized actions for SBC Staff, Maximus and SBC Contact Centre (Mihai helped confirm)
- added public_user role
- updated unit tests


bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-130-fqvysi29.web.app

web/business-registry-dashboard/app/utils/authorizations.ts
```diff
@@ -57,7 +74,7 @@ const BusinessRegistryStaffRoles = [
   AuthorizedActions.ADD_ENTITY_NO_AUTHENTICATION,
   AuthorizedActions.STAFF_DASHBOARD,
   AuthorizedActions.MANAGE_OTHER_ORGANIZATION,
-  AuthorizedActions.RESTORE_OR_REINSTATE
+  AuthorizedActions.RESTORATION_REINSTATEMENT_FILING
```
severinbeauvais, line 77: Is there any other special NR handling for staff only?

web/business-registry-dashboard/app/utils/authorizations.ts
```diff
@@ -79,12 +96,11 @@ const ContactCentreStaffRoles = [] as AuthorizedActions[]
...
 ]
 
 /**
- * The roles if the user is none of the other types.
- * Ultimately we won't need this list and we'll just check auth roles for everything.
+ * The authorized actions if user is Public User.
+ * Ultimately we won't need this array and we'll just fetch these.
  */
-const DefaultRoles = [] as AuthorizedActions[]
+const PublicUserActions = [] as AuthorizedActions[]
```
severinbeauvais, line 106: This is correct -- no special actions -- but it looks odd. Should we add a special actions that all roles would have, like VIEW_BRD?

JazzarKarim: It does look odd but I don't think that action would fit in. All users are going to be able to view the BRD right? Why would we have that action in this case?

JazzarKarim: I'm going to scour the repo and see if there are any special actions that would make sense to have for public users.

severinbeauvais: I was thinking about having an action that might be authorized for everyone.

I can't think of a user who can't see their account's BRD, or can't action most items in the list.

I agree with you but it still looks weird to have roles without actions 🤷‍♂️

JazzarKarim: I could add actions like:
- SEARCH_BUSINESSES
- VIEW_BUSINESS_INFORMATION

But I feel like they're going to be there just to be there. At the end of the day, all user types will be able to do this. I'm on the fence of not adding these and keeping things as is. Thoughts?

web/business-registry-dashboard/app/enums/authorized-actions.ts
```diff
@@ -4,7 +4,12 @@
  */
 export enum AuthorizedActions {
   ADD_ENTITY_NO_AUTHENTICATION,
+  HANDLE_BUSINESSES,
+  HANDLE_NAME_REQUESTS,
+  HANDLE_SOCIETIES,
+  HANDLE_TEMPORARY_BUSINESSES,
```
severinbeauvais, line 10: Can we come up with a verb other than Handle?

What is the actual action name, "manage something"?

web/business-registry-dashboard/app/utils/authorizations.ts
severinbeauvais: See my other comment. Resolving this one.

severinbeauvais: Well, I slept on it and reviewed your changes and re-read your comments above, and I'm on the fence.

I don't currently know of a case where we'd allow a user to view the BRD but not action anything, but it's possible later.

What does SEARCH_BUSINESSES protect? The ability to search for a business or NR to affiliate?

Should deleting an affiliation be protected?

web/business-registry-dashboard/app/enums/authorized-actions.ts
severinbeauvais: MANAGE_BUSINESS
MANAGE_NR
CREATE_DRAFT
SEARCH_BUSINESS
SEARCH_NR (or combine these 2)

JazzarKarim: Makes sense, changed!

web/business-registry-dashboard/app/enums/authorized-actions.ts
```diff
@@ -4,12 +4,12 @@
...
-  HANDLE_SOCIETIES,
-  HANDLE_TEMPORARY_BUSINESSES,
+  MANAGE_BUSINESS,
+  MANAGE_NR,
+  MANAGE_SOCIETY,
+  RESUME_DRAFT,
   MANAGE_OTHER_ORGANIZATION,
   RESTORATION_REINSTATEMENT_FILING,
-  SEARCH_BUSINESSES,
+  SEARCH_ENTITIES,
```
severinbeauvais, line 13: A NR isn't an entity. I don't mind SEARCH_BUSINESS_NR or maybe you have a better idea :)

web/business-registry-dashboard/app/components/Table/AffiliatedEntity/Action.vue
```diff
@@ -49,13 +49,13 @@ const checkPrimaryActionAuthorization = (item: Business): boolean => {
   let requiredAction: AuthorizedActions
 
   if (isTemporaryBusiness(item)) {
-    requiredAction = AuthorizedActions.HANDLE_TEMPORARY_BUSINESSES
+    requiredAction = AuthorizedActions.RESUME_DRAFT
   } else if (isNameRequest(item)) {
-    requiredAction = AuthorizedActions.HANDLE_NAME_REQUESTS
+    requiredAction = AuthorizedActions.MANAGE_NR
```
severinbeauvais, line 54: Is this to open the NR or "use it"?

ArwenQin (commented):
Looks good!
Waiting for decisions on the action buttons.

web/business-registry-dashboard/app/enums/authorized-actions.ts
JazzarKarim: This action handles both the usage of the NR and opening it in Namerequest UI. So, they're already combined 🙏

severinbeauvais: Sounds good. Resolving.

web/business-registry-dashboard/app/components/Table/AffiliatedEntity/Action.vue
```diff
@@ -41,6 +41,31 @@ const emit = defineEmits<{
 const invalidStatuses = [AffiliationInvitationStatus.Pending, AffiliationInvitationStatus.Expired, AffiliationInvitationStatus.Failed]
 const isButtonActionProcessing = ref(false)
 
+/**
+ * Check authorization for primary action based on type
+ * Returns true if authorized, false and shows error modal if not authorized
```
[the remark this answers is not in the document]
JazzarKarim: As per my conversation with Jacqueline, we're shying away from disabling/hiding buttons.

We're keeping that dialog but updating the text. Please see latest commit. Thanks!

web/business-registry-dashboard/app/locales/en-CA.ts
```diff
@@ -218,8 +218,8 @@ export default {
       description2: 'Make sure you entered the link correctly, as errors in the link may cause this message to appear.'
     },
     notAuthorized: {
```
severinbeauvais, line 220: Is it worth changing the label to match the title?

web/business-registry-dashboard/app/components/Table/AffiliatedEntity/Action.vue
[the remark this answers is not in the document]
severinbeauvais: No problem. Resolving.

web/business-registry-dashboard/app/locales/en-CA.ts
JazzarKarim: I thought about it but in my head, since we're using isAuthorized conditional in the code, it better resembles what the dialog is although the title is different. I'm good for both.

severinbeauvais (approved):

web/business-registry-dashboard/app/locales/en-CA.ts
severinbeauvais: Haha, I was going to agree with you.

The new name will make it easier to find 😃

PR #131, opened 2025-06-04
28805 - Extra fixes and design changes for filtering and pagination

JazzarKarim:
https://github.com/bcgov/entity/issues/28805

~need to confirm design with Jacqueline~ Design to be implemented: https://www.figma.com/design/6VCttiaobFtKqZPecb9Saj/Business-Dashboard?node-id=256-17836


severinbeauvais (approved):
Looks OK to me.

web/business-registry-dashboard/app/stores/affiliations.ts
```diff
@@ -240,6 +240,10 @@ export const useAffiliationsStore = defineStore('brd-affiliations-store', () =>
           // Add each status as a separate query parameter
           // This will output: ?status=Active&status=Expired&status=Approved
           affiliations.filters.status.forEach((status) => {
+            // Special case for awaiting review to map to NE since it's already defined in Namex
+            if (status === EntityStateStatus.AWAITING_REVIEW) {
+              status = NrState.NE
```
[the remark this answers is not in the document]
severinbeauvais: Hmm, that's weird.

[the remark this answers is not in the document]
JazzarKarim: @Rajandeep98 Would you be able to elaborate here by any chance please? This is expected? Both the status=Draft and status=NE show Pending Staff Review NRs.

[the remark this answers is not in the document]
severinbeauvais: I see new code here. Resolving this conversation.

JazzarKarim:
Just a heads up, updated the description with the new design. Working on implementing that...

web/business-registry-dashboard/app/components/Table/AffiliatedEntity/index.vue
```diff
@@ -246,6 +246,13 @@ const mapDetailsWithEffectiveDate = (details: any[], row: any) => {
                 />
                 <span>{{ option }}</span>
               </div>
+              <!-- Add divider before specific options to separate groups -->
+              <template v-if="option === FilterTypes.NAME_REQUEST || option === FilterTypes.INCORPORATION_APPLICATION">
+                <div
+                  data-divider
+                  class="absolute right-0 w-full cursor-pointer border-t border-bcGovGray-300 py-5"
+                />
+              </template>
```
JazzarKarim, line 255: This took way longer that I thought it would to implement. I was running into all sorts of issues since putting a divider inside the USelectMenu made it selectable as well. This here now looks good and works well (after trying all sorts of different solutions).

web/business-registry-dashboard/app/components/Table/AffiliatedEntity/index.vue
```diff
@@ -214,12 +214,12 @@ const mapDetailsWithEffectiveDate = (details: any[], row: any) => {
             :options="affStore.typeOptions"
             selected-icon="hidden"
             multiple
-            :ui="{ trigger: 'flex items-center w-full h-[42px]' }"
+            :ui="{ trigger: 'flex items-center w-full h-[42px] min-w-[210px]' }"
```
JazzarKarim, line 217: This here is kind of a QoL fix. The type filter drop down looked weird on select or deselect of options. It should look much better now.

web/business-registry-dashboard/app/components/Table/AffiliatedEntity/index.vue
```diff
@@ -246,6 +246,16 @@ const mapDetailsWithEffectiveDate = (details: any[], row: any) => {
...
+              <!-- Add divider before specific options to separate groups -->
+              <template
+                v-if="affStore.enablePagination &&
+                  (option === FilterTypes.NAME_REQUEST || option === FilterTypes.INCORPORATION_APPLICATION)"
+              >
+                <div
+                  data-divider
+                  class="absolute right-0 w-full cursor-pointer border-t border-bcGovGray-300 py-5"
+                />
+              </template>
```
JazzarKarim, line 258: This took way longer that I thought it would to implement. I was running into all sorts of issues since putting a divider inside the USelectMenu made it selectable as well. This here now looks good and works well (after trying all sorts of different solutions).

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-131-7toij2yx.web.app

web/business-registry-dashboard/app/components/Table/AffiliatedEntity/index.vue
severinbeauvais: Looks good.

web/business-registry-dashboard/app/components/Table/AffiliatedEntity/index.vue
severinbeauvais: The Type dropdown looks good to me right now.

I still see some horizontal jumping around when I toggle filter options because the columns widths are dynamic. Is there any way to improve that? (low priority)

JazzarKarim: You're right. I can't seem to fix this fully/completely without breaking something else. I'll create a ticket for this, is that OK?

JazzarKarim:
> Is there any way of disabling the pagination instead of hiding it? It jumps around a bit when it's hidding/displayed. (low priority)

Done. Agreed.

> Also, it looks like the filtering isn't actually disabling: I can still change filters while the table is loading. (~medium or high priority because it can hammer the API if the user keeps changing filters while the previous table is still loading~ WAIT. You don't send new API requests while one is in flight, right? You just re-fetch if I changed the filters while the table is loading? If so then I think it's fine; no concerns about too many fetches.)

OK, so I disabled both the filtering and pagination while table is loading. That second part is true, but I still think it'd be nicer if we disabled those while table is loading. Provides a better user experience.

Rajandeep98 (approved):
lgtm

web/business-registry-dashboard/app/enums/filter-statuses.ts
```diff
@@ -0,0 +1,17 @@
+export const STATUS_FILTER_OPTIONS = [
```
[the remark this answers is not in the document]
severinbeauvais: Perfect, thanks!

web/business-registry-dashboard/app/components/Table/AffiliatedEntity/index.vue
JazzarKarim: What you said makes perfect sense to me. I want to run this by Jacqueline first though as this design has been passed by the design folks including Scott which is the design lead. So, I'm thinking, they definitely had something in mind there to have them grouped together like this.

severinbeauvais (approved):
I think it's a good user experience with the controls disabled during load. And right now, it's fast enough that it's not frustrating to wait for loading before I can click something else.

(Any changes to grouping can be future.)

Approved.

PR #136, opened 2025-06-20
29067 prevent search not finalized

JazzarKarim:
https://github.com/bcgov/entity/issues/29067

This was interesting to work through. 

- I had to figure out first how can I tell if this user is subscribed / has access to the BRD. That's done using the /products endpoint in Auth API
- After that, I had to figure out which params do we need and what are the values that I'll be using
- After getting that, I had to set into the store the value of the boolean. The boolean `subscriptionStatus` is my hero here 🦸 . It says whether the user is actually subscribed. 
- An unfinished account will not have an active subscription.
- After setting into the store on load, I can now use that everywhere.
- Created the dialog. Exiting the dialog will go back as mentioned in the ticket, per Mihai's request
image
- Fixed unit tests


bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-136-opxnantn.web.app

severinbeauvais (approved):
Makes sense and looks good to me.

web/business-registry-dashboard/app/enums/org-product-codes.ts
```diff
@@ -0,0 +1,15 @@
...
+  ESRA = 'ESRA',
+  MHR = 'MHR',
+  MHR_QSHD = 'MHR_QSHD',
+  MHR_QSHM = 'MHR_QSHM',
+  MHR_QSLN = 'MHR_QSLN',
+  PPR = 'PPR',
+  RPT = 'RPT',
+  STRR = 'STRR',
+  VS = 'VS'
+}
```
severinbeauvais, line 15: This may go out of date and you only use a single item from this list. What do you think of declaring just BUSINESS (as a const or readonly) where you use it instead?

JazzarKarim: True and I agree. I had that first but then I decided to extract these guys from there auth web side of things. I'll take this out.

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-136-opxnantn.web.app

web/business-registry-dashboard/app/utils/account.ts
```diff
@@ -24,7 +24,7 @@ export async function hasActiveBusinessRegistryDashboardSubscription (orgId: num
     const parsedProducts = JSON.parse(products) as OrgProduct[]
 
     // Find the BUSINESS product (BRD)
-    const businessProduct = parsedProducts.find(parsedProduct => parsedProduct.code === OrgProductCodes.BUSINESS)
+    const businessProduct = parsedProducts.find(parsedProduct => parsedProduct.code === 'BUSINESS')
```
severinbeauvais, line 27: How about extracting this into a single const/readonly string (that you can describe with a comment)?

severinbeauvais: Looks great,

bobdev-94 (approved):

PR #139, opened 2025-06-24
28845 + 29226 - Fixed BRD Caching + UXA fixes

JazzarKarim:
https://github.com/bcgov/entity/issues/28845 and https://github.com/bcgov/entity/issues/29226

~I implemented a cache manage system. I couldn't reproduce this issue no matter what I tried. I'm kind of going for a leap of faith here 🦅 I'll be asking team members to keep on monitoring after this is merged and hopefully we don't see this issue anymore.~

~Step by step guide on what's happening here:~
- ~Added client-side plugin that runs on app load~
- ~Stores current app version in sessionStorage~
- ~Compares stored version with runtime config version~
- ~Automatically reloads page when version mismatch detected~
- ~Provides console logging for debugging version issues~

This thing won't work. If a cached version is showing, current will be the same as stored. I'm basically not doing anything.

I added aggressive no-cache headers. I don't think the issue is going to be there anymore. I'd like to move this forward as I spent more time than I hoped it would take. After this PR is merged, I'll be on the lookout to see if this happens again. I'll also ask the team to let me know if they see this issue again. 

Implemented new design for error dialog modals: https://www.figma.com/design/6MBqd7QAijpl6OXLvLXmos/Errors?node-id=313-70

~To put it simply. I am a user. I am running version 1.0.0. A deployment was done and version is now 1.0.1. I refresh. The browser will check the sessionStorage. Oh hey, 1.0.0 !== 1.0.1. Reload.~

### DIFF
--- a/web/business-registry-dashboard/package.json
+++ b/web/business-registry-dashboard/package.json
@@ -2,7 +2,7 @@
   "name": "business-registry-dashboard",
   "private": true,
   "type": "module",
-  "version": "1.1.14",
+  "version": "1.1.15",
   "scripts": {
     "build-check": "nuxt build",
     "build": "nuxt generate",


bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-139-aikh7cdo.web.app

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-139-aikh7cdo.web.app

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-139-aikh7cdo.web.app

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-139-aikh7cdo.web.app

web/business-registry-dashboard/firebase.json
```diff
@@ -24,6 +24,7 @@
             },
             { "key": "Cache-Control", "value": "no-cache, no-store, must-revalidate"},
             { "key": "Pragma", "value": "no-cache"},
+            { "key": "Expires", "value": "0"},
```
severinbeauvais, line 27: I'm not sure, but does this mean the content never expires?

web/business-registry-dashboard/nuxt.config.ts
```diff
@@ -3,13 +3,33 @@ export default defineNuxtConfig({
...
+    },
+    // Add aggressive no-cache for all routes
+    routeRules: {
+      '/**': {
+        headers: {
+          'Cache-Control': 'no-cache, no-store, must-revalidate',
+          Pragma: 'no-cache',
+          Expires: '0'
+        }
+      }
```
severinbeauvais, line 32: I'm not sure about Firebase configuration, but does this prevent caching altogether?

Ideally, the client would only fetch the index.html file (example below). If the content of it hasn't changed then the cached version of the rest of the modules can be used. If the index.html file has changed then everything needs to be reloaded. Does this config work like this?

image

severinbeauvais (commented):
I haven't learned the Firebase config yet, but Google returns many helpful-looking results...

https://www.google.com/search?q=firebase+cache+busting

https://firebase.google.com/docs/hosting/manage-cache

https://nuxt.com/deploy/firebase

web/business-registry-dashboard/firebase.json
severinbeauvais: I'm kinda hoping @pwei (or someone else with web server experience) can help set this up optimally.

web/business-registry-dashboard/nuxt.config.ts
JazzarKarim: You're right about this. Let me know what you think about latest commit. Basically now, pages now are not cached but static values are. Thoughts?

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-139-aikh7cdo.web.app

web/business-registry-dashboard/firebase.json
JazzarKarim: I don't see it happening even without this code here. I implemented selective caching. Basically:
First visit: Downloads everything (same as before)
Return visits: Only downloads index.html (~5KB) if nothing changed
After updates: Only downloads changed files, reuses cached assets that haven't changed

web/business-registry-dashboard/nuxt.config.ts
severinbeauvais: This looks great, but how did you test that chunks are being cached?

I'm trying the preview URL and it looks like it's fetching everything (except some SVG images) every time.

JazzarKarim: I'm trying something out in my latest commit.

For some reason, no matter what I try, I can't seem to get these chunks to be cached 😞 

The idea is, I'm trying to change the Cache-Control header here but to no avail:
image

bcregistry-sre:
Temporary Url for review: https://business-registry-dev--pr-139-aikh7cdo.web.app

web/business-registry-dashboard/nuxt.config.ts
severinbeauvais: It didn't make any difference for me when I unset "Disable cache" here. FWIW.

image

deetz99 (approved):
👍

severinbeauvais (commented):
What I'm seeing now is a lot of HTTP 304 (meaning the requested resource has not changed since the last time it was accesses by the client, so the client can use its cached version).

So, it looks like this is working for (most) of the calls (but, strangely, not some of the static HTML pages).

image

Should something similar be done for other Nuxt UIs (like Business Dashboard)? I suggest a new ticket be created for that, and that this solution be shared with other teams so they can create tickets for their own UIs as needed.

PR #144, opened 2025-06-26
28949 - officers save and resume

deetz99:
issue: bcgov/entity#28949

- refactor error modals to use common component, will map text content based on error code matching in i18n file
- implement legal api methods:
  - create and update draft filings
  - fetch business tasks
  - fetch and validate draft filing content
- add types/enums to support legal api filings
- update page to save, save and resume and submit draft filings
- update store to load draft filing and validate that a changeOfOfficers filing can actually be submitted on initial load instead of checking on submit
- validate draft util to check if the filing header status === 'DRAFT' and if the expected filing name object key is in the response


bcregistry-sre:
Temporary Url for review: https://business-people-dev--pr-144-7qerx37n.web.app

PR #146, opened 2025-06-30
29252 - Fix NR Conditionally approved cannot be used

JazzarKarim:
https://github.com/bcgov/entity/issues/29252

I could not test this because of DEV issues. NRs are not showing. I'm kind of going blindfolded here but most probably this is going to work.

UDATE: I actually tested this in TEST. I'm certain this will fix the issue.


deetz99 (approved):
👍

PR #159, opened 2025-07-07
29399 - officers - folio input

deetz99:
*Issue #:* bcgov/entity#29399

**Correct design still TBD**

*Description of changes:*
- update filing payload to be more generic to accept folio number
- update filing types to match new structure
- add folio number and filing header type
- update save/submit methods to use new structure
- add placeholder input on page for folio number
- update tests


thorwolpert (approved):
with the folio in, is it a wrap?

deetz99:
> with the folio in, is it a wrap?

@thorwolpert 

The input field is just temporary so it'll need to be updated with the correct design once it's created. But as far as logic goes I believe so.

PR #168, opened 2025-07-15
add read only version of shares/series

BrandonSharratt:
*Issue #:* /bcgov/entity#28829

*Description of changes:*
Added support for Connect Api
Fetch share classes for the business from it
Render Share classes and Series in a read only version of the table (has buttons but they just do a console log)

### DIFF
--- a/web/post-restoration-transition-application/.env.sample
+++ b/web/post-restoration-transition-application/.env.sample
@@ -1,0 +1,46 @@
+# canada post
+NUXT_ADDRESS_COMPLETE_KEY=""
+
+# vaults keycloak
+NUXT_KEYCLOAK_AUTH_URL=""
+NUXT_KEYCLOAK_REALM=""
+NUXT_KEYCLOAK_CLIENTID=""
+
+# auth api
+NUXT_AUTH_API_URL=""
+NUXT_AUTH_API_VERSION="/api/v1"
+# pay API
+NUXT_PAY_API_URL="https://pay-api-dev.apps.silver.devops.gov.bc.ca"
+NUXT_PAY_API_VERSION="/api/v1"
+# status API
+NUXT_STATUS_API_URL="https://status-api-dev.apps.gold.devops.gov.bc.ca"
+NUXT_STATUS_API_VERSION="/api/v1"
+# legal API
+NUXT_LEGAL_API_URL="https://legal-api-dev.apps.silver.devops.gov.bc.ca"
+NUXT_LEGAL_API_VERSION="/api/v2"
+
+# web urls
+NUXT_AUTH_WEB_URL="https://dev.account.bcregistry.gov.bc.ca/"
+# NUXT_BASE_URL="https://dev.bcregistry.gov.bc.ca/"
+NUXT_BASE_URL="http://localhost:3000/"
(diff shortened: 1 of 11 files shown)


web/post-restoration-transition-application/i18n/locales/en-CA.ts
```diff
@@ -67,5 +68,24 @@ export default {
         TRANP: 'Post Restoration Transition Application'
       }
     }
+  },
+  shares: {
```
kialj876, line 72: we're trying to keep this file as simple as possible. I think all of these can just go under 'label' like everything else instead of its own nested json for shares. The only reason we have 'ConnectFeeWidget' like this is because its a connect component (and we're planning to flatten that later)

web/post-restoration-transition-application/app/plugins/connect-api.ts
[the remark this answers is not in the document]
BrandonSharratt: it's very similar, except it's using a different url / config option, I didn't see any that supported it out of the box which is where share structure info comes from

web/post-restoration-transition-application/app/components/Shares/Index.vue
```diff
@@ -0,0 +1,166 @@
...
+                console.log('Delete', row)
+            }
+        }
+    ]
+}
+
+</script>
+
+<template>
+    <h1>{{ title }}</h1>
```
hfekete, line 131: This is a constant ? Always to be "shares" ?

web/post-restoration-transition-application/.env.sample
```diff
@@ -0,0 +1,46 @@
...
+# gtm/analytics
+# NUXT_GTAG_ID="G-GKRC2V8PT4"
+# NUXT_GTAG_DEBUG="true"
+# NUXT_GTM_ID="GTM-559DC9S4"
+
+# dev env
+PLAYWRIGHT_TEST_BCSC_USERNAME=""
+PLAYWRIGHT_TEST_BCSC_PASSWORD=""
+
+NUXT_CONNECT_API = "https://test.api.connect.gov.bc.ca/business-dev"
```
hfekete, line 44: I think this is just legal-api endpoint but they have this weird link building currently in GCP dev.

web/post-restoration-transition-application/app/plugins/connect-api.ts
```diff
@@ -0,0 +1,38 @@
+export default defineNuxtPlugin(() => {
+  const config = useRuntimeConfig()
+  const { $keycloak } = useNuxtApp()
+  const appName = config.public.appName
+
+  const api = $fetch.create({
```
hfekete, line 6: have you tried using `useLegalApi` composable for this fetch ?

web/post-restoration-transition-application/pnpm-lock.yaml
```diff
@@ -1,6224 +1,1003 @@
-lockfileVersion: '9.0'
+lockfileVersion: '6.0'
```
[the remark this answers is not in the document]
BrandonSharratt: fixed

BrandonSharratt:
This PR is made redundant by #174

PR #174, opened 2025-07-22
Edit shares

BrandonSharratt:
*Issue #:* /bcgov/entity#29294

This PR also includes the changes for #168 and additional changes making the other one obsolete (so this also includes the changes for /bcgov/entity#28829 )

*Description of changes:*
Added validation for shares, and all actions -- move up, move down, remove, edit, add
The button for add series exists but doesn't yet do anything.


web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
```diff
@@ -0,0 +1,298 @@
+<script setup lang="ts">
+import currencySymbolMap from 'currency-symbol-map/map'
+import type * as z from 'zod'
+
+const t = useNuxtApp().$i18n.t
+const filingStore = usePostRestorationTransitionApplicationStore()
+const {
+  shareClasses,
+  editingShareIndex
+} = storeToRefs(filingStore)
```
hfekete, line 10: can you add one additional tracking option (also affected by undo if we have it).
The option to track is "special rights and permissions" for the share class

this is for the reason if there were changes to that part, we require them to add article/court order date.

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
```diff
@@ -0,0 +1,298 @@
...
+
+const shareName = ref<string>(
+  shareValues?.value?.name.substring(0,
+                                     shareValues?.value?.name?.length - SHARES_TEXT.length)
+  || ''
+)
+
+const clickMaxShares = () => {
+  shareValues.value.hasMaximumShares = true
+  hasMaxShares.value = ''
```
hfekete, line 81: why have `hasNoMaxShares` and `hasMaxShares`? would it not make sense to have only one ? as they need to be opposite, no ?

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
```diff
@@ -0,0 +1,298 @@
...
+
+const clickNoMaxShares = () => {
+  shareValues.value.hasMaximumShares = false
+  hasMaxShares.value = 'false'
+  hasNoMaxShares.value = t('label.noMax')
+}
+
+const clickParValue = () => {
+  shareValues.value.hasParValue = true
+  hasParValue.value = ''
```
hfekete, line 93: same as for has(No)MaxShares, do we need both ?

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
```diff
@@ -0,0 +1,298 @@
...
+      }
+      return
+    }
+    shareValues.value.name = shareName.value + SHARES_TEXT
+    if (editingShareIndex.value !== -1) {
+      shareClasses.value[editingShareIndex.value] = shareValues.value
+    } else {
+      shareClasses.value.push(shareValues.value)
+    }
+    resetData()
```
hfekete, line 147: why reset data ?

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
```diff
@@ -0,0 +1,298 @@
...
+  }
+}
+
+const cleanData = () => {
+  if (shareValues.value.hasParValue === false) {
+    shareValues.value.currency = null
+    shareValues.value.parValue = null
+  } else {
+    const val: number = parseInt(shareValues.value.parValue) || 0
+    if (val == shareValues.value.parValue) {
```
hfekete, line 161: what is this line for ? 

you can do:
`const val: number|undefined = parseInt(shareValues.value.parValue) || undefined`
and even better:
`const val: number = parseInt(shareValues.value.parValue)`
then check with this function: `isNaN(val)`

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
```diff
@@ -0,0 +1,298 @@
...
+
+        <div class="flex">
+          <URadioGroup
+            v-model="hasMaxShares"
+            :items="['']"
+            class="flex-0 mr-3 align-bottom text-base"
+            :ui="{
+              container: 'text-base h-[56px]'
+            }"
+            @change="clickMaxShares()"
```
hfekete, line 213: if you are triggering function on change, lets not call it clickX rather lets call it xHandler.
for this example:
`maxSharesChangeHandler`
you could do same for click, calling it handler after.

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
```diff
@@ -0,0 +1,298 @@
...
+                class="h-[56px] w-full"
+              />
+            </UFormField>
+          </div>
+        </div>
+        <div>
+          <URadioGroup
+            v-model="hasNoParValue"
+            :items="[$t('label.noPar')]"
+            @change="clickNoParValue()"
```
hfekete, line 273: same here for the name

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
```diff
@@ -0,0 +1,298 @@
...
+
+        <div class="flex">
+          <URadioGroup
+            v-model="hasParValue"
+            :items="['']"
+            class="flex-0 mr-3 align-bottom text-base"
+            :ui="{
+              container: 'text-base h-[56px]'
+            }"
+            @change="clickParValue()"
```
hfekete, line 242: same here for the name

web/post-restoration-transition-application/app/components/Shares/Index.vue
```diff
@@ -0,0 +1,287 @@
...
+  if (addingShare.value) {
+    return
+  }
+
+  if (anyExpanded.value && editingShareIndex.value !== row.index) {
+    return
+  }
+
+  anyExpanded.value = true
+  if (row.getIsExpanded()) {
```
hfekete, line 143: why set anyExpanded to false when row.getIsExpanded is true ?

web/post-restoration-transition-application/app/components/Shares/Index.vue
```diff
@@ -0,0 +1,287 @@
...
+    <UButton
+      icon="i-mdi-plus"
+      :label="$t('label.addShare')"
+      color="primary"
+      class="inline-block my-4 px-5 py-3"
+      variant="outline"
+      :ui="{
+        label: 'align-top'
+      }"
+      aria-label="add share"
```
hfekete, line 205: aria label can be same as label, lang value ?

web/post-restoration-transition-application/app/interfaces/shares.ts
```diff
@@ -0,0 +1,65 @@
+import * as z from 'zod'
+
+export interface Series {
+    id: number | null // can be null when adding
```
hfekete, line 4: can we use undefined instead of null 
and for values below?

web/post-restoration-transition-application/i18n/locales/en-CA.ts
```diff
@@ -44,7 +45,36 @@ export default {
     resolutionOr: 'Resolution or',
     save: 'Save',
     sameAsMailingAddress: 'Same as Mailing Address',
-    to: 'to'
```
hfekete, line 47: I'll have to bother you if you can add texts in alphabetical order :P

We have a note on top, that might not be perfect, you can improve it that says use alphabetical unless otherwise noted. This refers to order of the text keys inside their respective place.

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
```diff
@@ -0,0 +1,298 @@
+<script setup lang="ts">
+import currencySymbolMap from 'currency-symbol-map/map'
+import type * as z from 'zod'
```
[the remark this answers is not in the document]
BrandonSharratt: Done

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
BrandonSharratt: Due to the design and the functionality of RadioGroup they need to have different models. I don't like it either but they can't be the same model/group because of the text fields that become required

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
BrandonSharratt: So that after you add, if you immediately add another it doesn't prepopulate with the same data (it reuses the same component)

web/post-restoration-transition-application/app/components/Shares/Index.vue
BrandonSharratt: So this function does both close and open. initially sets expanded to true and then checks what it's doing, if the row is already expanded then the toggle at the end of the function will close it so we set it to false, as you're closing the only one thats open. If it's going to be open we set the editing index to the row that we're editing

web/post-restoration-transition-application/app/components/Shares/Index.vue
BrandonSharratt: done

web/post-restoration-transition-application/app/interfaces/shares.ts
BrandonSharratt: Modified, I'll have to recheck all validations as they depended on null with zod

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
hfekete: Thanks for adding it

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
hfekete: I am happy this could been resolved with one variable ! :D

web/post-restoration-transition-application/app/components/Shares/AddEdit.vue
```diff
@@ -0,0 +1,310 @@
...
+  )
+)
+
+const hasNoMaxShares = ref<string>(shareValues.value.hasMaximumShares ? '' : t('label.noMax'))
+const hasNoParValue = ref<string>(shareValues.value.hasParValue ? '' : t('label.noPar'))
+const errors = ref<ZodError[]>([])
+
+const shareName = ref<string>(
+  shareValues?.value?.name.substring(0,
+                                     shareValues?.value?.name?.length - SHARES_TEXT.length)
```
hfekete, line 72: Haha, I think you might be able to move sharesValues?... into above line without triggering the line length linter, but not a mandatory thing 🎃

hfekete (approved):
Good work, thanks for changes!

PR #199, opened 2025-08-20
Transition Application - playwright working

kialj876:
*Issue #:* /bcgov/entity#30298

*Description of changes:*


kialj876:
Will fix the CI run on Monday - think its missing some env var or something like that. Passing locally

web/post-restoration-transition-application/tests/e2e/specs/restoration-application.spec.ts
```diff
@@ -1,19 +1,19 @@
-import { test, expect } from '@playwright/test'
```
hfekete, line 1: whats the reason for commenting all this out ?

kialj876: added a comment with the ticket number

BrandonSharratt (approved):

PR #212, opened 2025-08-28
Base layer ci

deetz99:
*Issue #:* bcgov/entity#29333

*Description of changes:*
- add test config to layer
- update env.example
- add base layer ci workflow
- e2e tests failing since its the example test from officers, but they are running in the checks


deetz99:
> good to see the playwright stuff being built out

Yeah just still need to sort out the auth/secrets stuff.

PR #220, opened 2025-09-03
Refactor person roles

deetz99:
*Issue #:* /bcgov/entity#29341

*Description of changes:*
- fully refactor person roles common stuff into business base layer

Lots of files but 90% is reorganizing from person-roles into base layer or deleted.

Main changes:
- legalApi -> businessApi
- useConnectDetailHeaderStore -> useBusinessTombstoneStore
- move several i18n keys into person roles
- useModal returns modal instance instead of helper methods
- pass any icon, badge or link props to tombstone options
- delete some stuff from connect layers (eg: zod schemas for address and tests)
- rename form layout to filing layout
- remove 'btn' prefix for i18n and use 'label' only
- move some person roles deps into layer
- move public runtime config keys into layer

**tests/auth api fetch TBD**


thorwolpert (approved):
That's a lot of file reorg. Nice.

PR #506, opened 2026-06-26
32006 Generate a new UUID for any share class with a null id as part of the draft filings formatter

vysakh-menon-aot:
*Issue #:* /bcgov/entity#32006

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-ui license (BSD 3-Clause).

### DIFF
--- a/.changeset/early-numbers-double.md
+++ b/.changeset/early-numbers-double.md
@@ -2,4 +2,4 @@
 "@sbc-connect/nuxt-business-base": minor
 ---
 
-allow null id in share chass formatter to support draft filings
+Generate a new UUID for any share class with a null id as part of the draft filings formatter.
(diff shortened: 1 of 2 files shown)


deetz99:
We rely on the ID in the [composable](https://github.com/bcgov/business-ui/blob/main/packages/layers/base/app/composables/useManageShareStructure.ts), does the edit/change priority functionality still work when the ID is null?

vysakh-menon-aot:
> We rely on the ID in the [composable](https://github.com/bcgov/business-ui/blob/main/packages/layers/base/app/composables/useManageShareStructure.ts), does the edit/change priority functionality still work when the ID is null?

Good catch!

deetz99 (approved):
LGTM 👍 

Can you please update the changeset though?

I think we can ignore the CI build for now, SRE is upgrading to pnpm 11 and there's some workspace/config changes needed which is why the build is failing currently.

doug-lovett (approved):

PR #511, opened 2026-07-03
34023 - fix layer typecheck

deetz99:
*Issue #:* /bcgov/entity#34023

*Description of changes:*
- fix typecheck in base layer
- fix several type issues
- add typecheck step to base layer CI

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-ui license (BSD 3-Clause).


packages/layers/base/app/components/Form/DocumentUpload/index.vue
deetz99, line 1: Not a huge fan of the `as any` but we'd need bigger changes to the document state definition which is out of scope for this ticket.

vysakh-menon-aot (approved):

doug-lovett (approved):

